PR #17, opened 2014-03-24
Temporary creative tab naming fix

DrummingFish:
This should fix the localisation problem with the Creative Tab naming.
A better solution needs to be done.


Dolu1990:
Oky :)

Dolu1990:
Ahhh nice ^^ That true the old pop up of creative tab sucks XD

PR #18, opened 2014-03-25
Grammar and wording fixes - Part 5

DrummingFish:
This is huge. I'm sorry if some recipes break because of this (they shouldn't) but this was needed.

In addition:
The way you do the recipes needs to be changed so it doesn't look for the name of the item but the object.
This allows you to change the name of the item without it affecting any recipes associated with it. You have actually half done this with the Copper armour and tools but everything else does not.

The recipes should also be in a separate class to make it much easier.

### DIFF
--- a/mods/eln/Eln.java
+++ b/mods/eln/Eln.java
@@ -247,7 +247,7 @@ import cpw.mods.fml.relauncher.Side;
 
 
 
-@Mod(modid = "Eln", name = "Electrical age", version = "BETA-1.1.1")
+@Mod(modid = "Eln", name = "Electrical Age", version = "BETA-1.1.1")
 @NetworkMod(clientSideRequired = true, serverSideRequired = false, channels = { "miaouMod" }, packetHandler = PacketHandler.class)
 public class Eln {
 	/*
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi ^^
The procedure to change name of things is ctrl + f  ->  remplace all   in eln.java, It's a legacy.
When you say about object name, it's the UnlocalizedName ?

I actualy check change :)

Dolu1990:
i can't see diffs XD i don't know why, In any case, if a item is not found (recipe name don't match item name) that make crash the game directly ^^
I merge that and fix mistake (there are some) 
i push a event to you when i have commit a fonctionnal eln.java with your change :)

Dolu1990:
Oky, i have push a working eln.java ^^

Following things was not coherente between recipe and itemDescriptor :

Lamp Socket B Projector
Wireless Signal Receiver
Overheating Protection
LED vuMeter

Are there name all right ?

DrummingFish:
By the object name I mean using something like: 
new ItemStack(eggIncubator)

rather than: 
findItemStack("50V Egg Incubator", 1)

And yes they are all right sorry about that :)

Dolu1990:
Ahhh, hmmm made like notch as made in Block.java ?
all item has one class ?
It's technicaly not possible because most of items/bocks share the same Item/Block instance.
I have do that to have less id used, and less constraint.
The alternative is to save the itemDescriptor referance and call on this instance newItemStack()
but i'm a slacking guy, and i have only do that only for very common items like cable ^^

DrummingFish:
Then couldn't you do something like:
new ItemStack(eggIncubator, 1, 0)

1 = the stack size
0 = the metadata

Dolu1990:
yes can you do that, but it's a little "overkill", i think. and there are no save of corresponding metadata.
Maybe it's bether to define a final static string that containe item name in only one place ? and use findItemStack(myItemName) 
a other way is that :  public ElectricalCableDescriptor lowVoltageCableDescriptor;
lowVoltageCableDescriptor.newItemStack()

What to you think about that ?

DrummingFish:
Yeah that could work. Whatever is best for you I'm only trying to help :)

PR #46, opened 2014-06-21
Update README.md

meelock:
well you dont have to do this but... [...s]


Dolu1990:
Yes we sucks in english XD sorry for that ^^

PR #177, opened 2015-01-09
UNTESTED move block/item registrations from init to preinit where they belong

mjevans:
I would have liked to have tested this prior to submitting a pull request, but I honestly can't see any sort of project or build environment configuration included in the repository.

This is a bulk move of the most obvious things from load (init) to preinit where they belong.  However as there is no build environment it is not economical for me to test this change.
## 

A bug exists when other projects (E.G. custom oregen from other mods like TerraFirmaCraft or the COFH world gen) try to generate your blocks in the world.  Due to your late registration of the blocks (they should be registered in preinit, not load AKA init) this issues an NPE when the ores are generated, causing world crashes.
## 

http://greyminecraftcoder.blogspot.com/2013/11/how-forge-starts-up-your-code.html  Elaborates on what types of actions/events should occur at each stage.

You can probably find more resources at these locations:
http://www.minecraftforge.net/wiki/Category:Tutorials
http://www.minecraftforge.net/forum/index.php?topic=14048.0  Topic: [TUTORIAL] Getting Started with ForgeGradle  (Read 81064 times)

http://terrafirmacraft.com/f/topic/6576-help-with-modding/?p=96927  A lovely example of a mod with a build environment already included and a fairly easy to follow (if reading carefully and step by step performing) set of directions.

### DIFF
--- a/mods/eln/Eln.java
+++ b/mods/eln/Eln.java
@@ -529,6 +529,172 @@ public class Eln {
 
 		config.save();
 
+
+
+        //computerCraftReady = Utils.isClassLoaded("dan200.computercraft.ComputerCraft");
+
+
+        eventChannel = NetworkRegistry.INSTANCE.newEventDrivenChannel(channelName);
+
+        simulator = new Simulator(0.05, 1 / electricalFrequancy, electricalInterSystemOverSampling, 1 / thermalFrequancy);
+        nodeManager = new NodeManager("caca");
+        ghostManager = new GhostManager("caca2");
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi,
Im patching that, and that seem to be working : D
incredible magic XD
release come today. then can you ckeck it's ok for you ?

Dolu1990:
For buildscript and all these things, i'm not a good user.
How other are doing usualy ? Put the forge main folder as root of the git repository with a good git ignore ?

PR #184, opened 2015-01-22
Latest commits to merge into new gradle branch

DrummingFish:
My latest commits weren't included in the new branch so I am just checking if my changes are okay to be merged.


Dolu1990:
Oky, tell me when it's ready ^^

DrummingFish:
I was hoping you would merge it. I'm not necessarily comfortable with merging and I don't want to be responsible for breaking something. But I can if you're too busy.

PR #206, opened 2015-03-16
Fix achievement triggering.

### LINKED ISSUE #186: Achievements don't trigger
I have implemented a couple achievements but have been unable to get them to trigger. The "if" statements are entered but the triggers don't work.

**ElnFMLEventsHandler.java**

``` java
@SubscribeEvent
public void onCraft(ItemCraftedEvent e) {
    // Doesn't seem to trigger.
    if (e.crafting == Eln.findItemStack("50V Macerator", 1)) {
        e.player.triggerAchievement(Achievements.craft50VMacerator);
    }
}
```

**ElnForgeEventsHandler.java**

``` java
@SubscribeEvent
public void openGuide(GuiOpenEvent e) {
    EntityPlayer player = Minecraft.getMinecraft().thePlayer;

    // Doesn't seem to trigger.
    if (e.gui instanceof Root) {
        player.triggerAchievement(Achievements.openGuide);
    }
}
```

bloxgate:
Looks like you were trying to trigger the wiki achievement on the client. Added a simpleChannel that tells the server to do it instead.
Closes #186, but seems to cause a new bug of having the second trigger with the first and have no message.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .classpath
 .project
 .settings
+.metadata
 bin
 build
 eclipse
--- a/src/main/java/mods/eln/Achievements.java
+++ b/src/main/java/mods/eln/Achievements.java
@@ -13,8 +13,8 @@ public class Achievements {
     public static AchievementPage achievementPageEln;
 
     public static void init() {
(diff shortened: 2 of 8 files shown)


DrummingFish:
Knew I'd missed something. I'll look through it more in depth and merge if everything checks out. Thanks!

PR #211, opened 2015-04-02
Fixes server side crashes

bloxgate:
Looks like this crashes the server now that the achievements can actually trigger.


bloxgate:
Commits for this have been added to my other pull request.

PR #244, opened 2015-06-08
Use units in Circuts simulation

serathius:
Using Primitives is bad.
Using Primitives in physics calculation where you operate on units is very bad.

This changes should be introduced step by step. I started by mods.eln.sim.mna package. 

This should prevent calculation bugs and help finding existing ones(calculation of bad values is impossible, for example now adding resistance to current).

This will help understand equations and make code more readable.
This will help with later refactorization of code.

At this stage other packages use .getValue() method which is ugly but you cant change everything at once.

Tested on my test world with most elements, and H bridge.

### DIFF
--- a/src/main/java/mods/eln/entity/ReplicatoCableAI.java
+++ b/src/main/java/mods/eln/entity/ReplicatoCableAI.java
@@ -15,6 +15,7 @@ import mods.eln.sim.IProcess;
 import mods.eln.sim.ITimeRemoverObserver;
 import mods.eln.sim.TimeRemover;
 import mods.eln.sim.mna.component.Resistor;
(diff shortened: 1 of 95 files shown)


serathius:
No comment?

Dolu1990:
Waiting you on skype to talk ^^
capons1785 I have add you (siara_91?)

cm0x4D:
We did something similar in our python based open source electro-thermal simulator (gridsim). Performance went down dramatically, so we had to remove the unit system from the simulation loop. 

Did you made some performance comparisons? I mean doing new for each mathematical operation in game/simulation loops is not a good idea generally in terms of performance.

PR #256, opened 2015-06-26
Applied Energistics ME Interface Terminal crash

Baughn:
You might not want the "autoformat" commit, but IDEA kept bugging me. And if you have a better idea of how to fix the ME issue... but this code at least avoids a crash.


Baughn:
Oh, right. The actual crash: http://sprunge.us/MXTg

bloxgate:
Whoops, looks like you can [blame me for that crash](https://github.com/Dolu1990/ElectricalAge/commit/03d4cf88373c53da29699277e3077e6cc293656e) :frowning:. Sorry.

Baughn:
I don't think it would have worked, even without that change. The problem is that all the machines are represented by the same class, TransparentNode, and it isn't fully initialised (with a descriptor) when AE tries to figure out what to call it...

I think it'd be cleaner design to have a different class for each machine, but what do I know? I'm still figuring this codebase out. :X

Incidentally, my code makes it show up as "mods.eln.sixnode.TransparentNode" or so. Which isn't ideal, but much better than the crash. I'm not sure what the base minecraft code might have made of it.

Dolu1990:
The reason of single class is : one year ago, in minecraft you had to reservate ID for each block. and it make lot of conflict when you use multiple modes.
By using single block classes, this problem was removed by only use fiew blocks id.

Now it's doesn't realy make sense. but it's like that ^^

In fact there is XXXElement, xxxRender, xxxDescriptor for each Eln machine.

Dolu1990:
Alala, migration of minecraft mode codes through minecraft version is an adventure ;)

Baughn:
Mind if I rework that design next, then? It'd be a nice exercise.

metc:
Thanks for your contribution @Baughn

Dolu1990:
Hmm the problem is compatibilities, if we migrate it could be hard to preserve map compatibilities.
It's a legacy, but i don't think it's a dramatic legacy. Of course, lot of things that i have implement could be better, but it's already working, so :)

maybe one things could be to create a little bit like universal electricity mod, a common API for realistic electric simulation.
Ahhh, life is hard XD

Baughn:
I thought of that, yes. It's why I think it'll be a good exercise—I'm not that familiar with Minecraft modding yet, but in this case it should be easy to tell whether or not I got it right.

Dolu1990:
Sorry, my english is broken, i don't understand your question XD
It's about reimplement things by a better way ? I don't realy like this idea if map compatibilities is broked.

There are some blocks https://github.com/Dolu1990/ElectricalAge/tree/1.7.2-MNA-gradle/src/main/java/mods/eln/transparentnode
and some other : https://github.com/Dolu1990/ElectricalAge/tree/1.7.2-MNA-gradle/src/main/java/mods/eln/sixnode

♪♫

Dolu1990:
It's just the fact that things are actualy hiden into a Electrical age abstraction layer to don't interact directly with minecraft (because minecraft interface change every time :( )

Baughn:
@Dolu1990 Understood. I'd want to keep that, as well as map compatibility, but the way the code is structured right now makes it pretty hard to make it interact properly with AE.

PR #258, opened 2015-06-26
Reformat everything, by IDEA rules

Baughn:
Since you accepted the last one—

I asked IDEA to reformat everything. This is what it came up with; there should be no functional changes. We both use IDEA, and it does simplify things to follow a formatting standard, not least merging patches.


Dolu1990:
Hi, For me it's good. from the moment that it doesn't wrap line if it's more than 80 char ^^

But the problem is that other people are currently working of some branch, it could be difficult to them to merge change after that no ?

But about the idea of reformating all is ok i agree, just a question of timing with other people :)

Baughn:
True. That's exactly why I think reformatting is a good idea, since once there's a common format merges becomes easier.

There might never be a good time to do such a change painlessly, though. I don't have any overview of what people are working on EA. (Should we get an IRC channel? Probably...)

If you do have such an overview, feel free to poke me whenever you think would be a good time.

PR #276, opened 2015-08-04
Slow down the solar panel updates. And fix some typos.

Baughn:
Everything outside of SolarPanelSlowProcess.java is IDEA-generated refactoring and should be perfectly safe.

What brought this on? Well, I have a solar power station with 256 solar panels in it, and noticed tick times shooting up. There's no particular reason these should update all that often, once per second is plenty.

Still need large solar panels, or network optimization of some kind, but it's a start. I'm no longer crowding 50ms.


aschmois:
With PR experience, I recommend you to re-do this PR and **only** fix the changes you mentioned, I see you changed a few class names and such and should not be inside a PR. The smaller the PR the better really.

Baughn:
A fair point. I'll do that, splitting out cleanup to further commits. Dolu can merge as little or as much of this branch as he wants.

PR #277, opened 2015-08-06
Add a resistor and thermistor

### LINKED ISSUE #265: Add resistors
For many purposes. Inrush current limiters, generation-shedding to ground, etc. etc...

They're the sole missing fundamental circuit element, and I will have it. How hard do you think it would be to add them? I'm willing to give it a try, though I haven't had much time to work on EA yet.

I'm imagining an UI with two slots. One for a cable, to select the voltage; another for coal dust, to select the resistance.

Baughn:
Resolves #265 

You should go over this commit with a fine-grained axe. I'm probably doing all sort of things wrong, but at least it seems to work here.

### DIFF
--- a/src/main/java/mods/eln/Eln.java
+++ b/src/main/java/mods/eln/Eln.java
@@ -96,6 +96,7 @@ import mods.eln.sixnode.modbusrtu.ModbusRtuDescriptor;
 import mods.eln.sixnode.modbusrtu.ModbusServer;
 import mods.eln.sixnode.powercapacitorsix.PowerCapacitorSixDescriptor;
 import mods.eln.sixnode.powerinductorsix.PowerInductorSixDescriptor;
+import mods.eln.sixnode.resistor.ResistorDescriptor;
 import mods.eln.sixnode.thermalcable.ThermalCableDescriptor;
 import mods.eln.sixnode.thermalsensor.ThermalSensorDescriptor;
 import mods.eln.sixnode.tutorialsign.TutorialSignDescriptor;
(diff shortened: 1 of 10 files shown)


Baughn:
And I figured it out. :dancer:

Baughn:
Seemed to work at first, but now the temperature of the resistor _and surrounding network_ shows as "NaNM degrees C". I have no idea why.

EDIT: I have some idea why. Let's see...
E2: I think that fixed it? It's working in production now, no problems that I can see.

Baughn:
A "signal resistor" makes no particular sense, so perhaps I should rename this.

Dolu1990:
Hi,
I'm actualy underwater, i will look contant soon :)

Baughn:
When you do get back...

Could you maybe explain to me the difference between a SixNode and a TransparentNode? I have no idea why some items are one and some are the other, which is a bad place for me to be.

Dolu1990:
tranparent nodes are blocky things like furnace, transformer

sixnode are element that could be 6 on a signel cube,(one on each side), like cable, probes, ect :)

Dolu1990:
Sorry for delay, I realy need to free some time, I will try this week-end :)

Dolu1990:
Hi,

All seem fine, Juste one thing in term of performance is not good :
The resistance value is updated each cycle by ThermistorProcess. It imply loot of processing in the simulation loop, You call the ThermistorProcess by a slow process, it's good (20 hz), but it still to fast. 
I will add a filter to don't change  the value by less than 1 %, or simething like that ^^.

Thanks for the commit : D

Dolu1990:
Hoo, it seem that the resistance of the thermistor fall when temperature rise, is it the wanter beavorial ?

Baughn:
That's intended behaviour, yes. The one thing I couldn't decide was whether it should be configurable via silicon dust or some such; the default value I picked was determined such that the thermistor will never explode, but rather fall off to zero resistance just before it would.

I'm using it for an inrush current limiter.

Baughn:
SlowProcesses get run every tick? :O

I expected them to be in some sense slow. I'll be looking forward to that commit.

Dolu1990:
Hmm, in reality, the greate majority of conductor => resistance rise with temperature.
Maybe we can do as following :
Actual dust => resistance rise with temperature
Redstone or silicon dust => resistance fall with temperature

Yes slow process get run every tick, Try to call the setR function only when there is a mismatch bigger than 1 % :)

PR #292, opened 2015-09-06
Powerpole

Baughn:
Hoping you can get to these PRs eventually..


Dolu1990:
Hi, i will try this week end :)

Dolu1990:
I don't know why, but i have some issue to setup the environment (eclipse and IDEA)
It's not due to your submit, don't worry, but which one you use ?  eclipse or idea, do you had some issue to setup them ?

(idea doesn't load texture properly)
(eclipse workspace is not generated XD)

Baughn:
I use IDEA. I've tried Eclipse, and IDEA is so much better it's not funny.

There was in fact some problem with loading textures there, which I worked around locally... here: http://stackoverflow.com/questions/27623326/minecraft-forge-not-loading-textures

Dolu1990:
Hoo ok thanks :)
I have try, but no success,

Did you just remplace 

sourceSets {
    main {
        java {
            srcDirs = ['src/main/java']
        }
        resources {
            srcDirs = ['src/main/resources']
        }
    }
}

by  

sourceSets {
    main { output.resourcesDir = output.classesDir }
}

and then runnning the following command ?
gradlew setupDecompWorkspace idea --refresh-dependencies

Baughn:
I didn't replace; I _added_ the sourceSets bit of the above.

But I'm not quite sure anymore. At some point I reverted the change (because it broke command-line ./gradlew runClient), yet IDEA kept on working, probably because I haven't refreshed dependencies...

Baughn:
Also, I just remembered that (according to one player) there's a potential crash bug in this code. I haven't been able to replicate it, so I forgot about it a little bit, but... yeah.

If it exists, it's probably somewhere in the integration of pole connection/destruction with the rest of the Eln network.

Dolu1990:
Thanks :)
Do you have a skype or something to chat ? my skype is capons1785

Baughn:
I don't. My only real way to do real-time chat is via IRC, where you can find me on irc.rizon.net #erisia.

Or just use https://madoka.brage.info/#irc.

Dolu1990:
It seem that every names are baned from this channel by default, it's is possible ? i get a (+b) error :)

Baughn:
No, that's... definitely not right. We get new people all the time.

Can you try with a normal IRC client? It's irc.rizon.net channel #erisia.

Baughn:
I found that crash bug. Give me a day or two, I'll get this into a somewhat less scary state.

Baughn:
I think I fixed it. No more crashes, at least from that source.

It crashed at InterSystemAbstraction.java:36, with a NPE, if there were three poles in a triangle. Not under any other circumstances that I could find. Squares were fine, as were triangles connected to a larger network, even just one extra pole.

I'm not sure if this really represents a bug in my code or not, as I don't understand that abstraction. I'm also not sure why 847705b would fix it. It's worthwhile anyway, just for the slightly lower CPU use, but... mmh. Right.

Dolu1990:
Hi :)
I have try the last week and this week to join the IRC, by using web IRC and client IRC, but alwas the same banning error, even with a registred nick XD
Is there another way to have a private chat ?

Baughn:
Hmm, it's possible that you were caught by one of the pre-existing bans. A bit unexpected, but possible. I've removed them; they shouldn't matter anymore.

Can you try again? And if #erisia just doesn't work, try #nanoha instead; I'm in that one as well.

Baughn:
Actually, never mind this pull request ATM. Or, merge it if you feel safe about that, but I'm going to rewrite it in Kotlin -- assuming I can get that to work with Forge -- and fix the class structure and dependencies while I'm at it.

It'll be awesome. You'll see.

CarolAnthony:
psst Just release into 1.8 my request powerpole pls :)
 it woud be nice :+1:

Baughn:
Having written a lot of code for this, I've pretty much decided that I need to _first_ fix the TileEntity confusion. For a lot of reasons, including code I haven't sent you yet.

So I'm closing this now, and will reopen later.

CarolAnthony:
Ok :D

PR #311, opened 2015-10-24
fix #242

### LINKED ISSUE #242: Diode blows up when connected to full voltage
I have two cases with same circuit for diode 10A and diode 25A(didn't test signal):
source(50V), diode conducting, receiver(lamp 125 ohm)
Predicted current 400mA

When I set 50V from 0 on source diode explodes.
When I set first 1V from 0, and later 50V circuit remains stable, with correct current.

marcin212:


Baughn:
So, all this does is give diodes a one-tick grace period when they switch from conducting to resisting.

I have some idea of why this might work, but can you explain it in your own words? Ideally, in the form of comments on your commit.

Dolu1990:
Hoo in fact, some other things have this kind of grace tick, like for example every over current.trigger.
So make this for diode can help :)

marcin212:
When you change the diode's resistance from a high to low, the voltage difference between the cathode and anode reaches max value, which causes it to emit very high amounts of energy in one tick, before the voltage gets recalculated.

PR #321, opened 2015-11-22
WAILA Integration

bloxgate:
Provides WAILA integration for TransparentNodes. At the moment it only displays for batteries. By default only displays charge, energy, and life, but the wailaEasy config option allows voltage and discharge current to be displayed too.
- New method to get a HashMap of strings to put into the WAILA body for transparent nodes: getWaila()
- WAILA on all TransparentNodes
- 1 new config option for "easy mode" on the WAILA display

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,8 +54,21 @@ configurations {
     compile.extendsFrom external
 }
 
+repositories {
+    mavenCentral()
+	maven { // The repo from which to get waila
+        name "Mobius Repo"
+        url "http://mobiusstrip.eu/maven"
+    }
+    maven { // the repo from which to get NEI and stuff 
+        name 'CB Repo'
+        url "http://chickenbones.net/maven/"
+    }
+}
+
 dependencies {
     external files("libs/commons-math3-3.3.jar", "libs/modbus4J.jar", "libs/seroUtils.jar")
+	compile  'mcp.mobius.waila:Waila:1.5.10_1.7.10'
 }
 
 jar {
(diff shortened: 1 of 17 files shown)


Baughn:
You have one commit implementing your original design, then another backing it out and replacing it with the new one...

I think you should rebase and squash this, possibly into a single commit. If you're feeling ambitious, then make it two: One for the base system, then another for the battery-specific implementation.

Oh, and if the only thing stopping getWaila from working on non-batteries is code explicitly telling it not to, then please remove that right away.

bloxgate:
@Baughn Done

Baughn:
Just squash these two commits. Again, you're writing code only to then back it out.

I assume you've tested that Eln still works without Waila installed?

bloxgate:
@Baughn The client should be unable to ask for a non-transparent node coordinate as the code to lookup the data for a coordinate is only called while looking at TransparentNodes. The exception is not being printed as it would generate a lot of console spam.

Baughn:
(You don't need to keep saying @Baughn. The issue is assigned to me, so I get notifications anyway.)

One of the basic rules of client-server coding is _don't trust the client_. In this case, even without deliberate hacks, it'd be fairly simple for the client and server to go out of sync; imagine if a block is broken by someone else, while you're looking at it.

bloxgate:
Either way, I added a check for it that doesn't send a message back to the client if its an invalid node. See da26dd3

Baughn:
I tested this, and it seems to work...

But now there are merge conflicts. :P

bloxgate:
Merge conflicts fixed.

PR #353, opened 2015-12-29
Assets #351

metc:
This PR is related to the issue https://github.com/Dolu1990/ElectricalAge/issues/351.

I created a resource white-list to include only necessary asset files in the release Jar, and I also detected some unused files.
I also added a `META-INF` file to the release Jar.

I tested this locally on the tuto map and I thinks every textures, model, sounds are OK. Many files have been removed from the release jar (r50 file was 6.3Mo, r51 is 5.7Mo).

@cm0x4D @lambdaShade @Dolu1990 please review. I will merge this if it is ok for you.

**From now, please use lowercase file extensions for asset files, or the files will not be filtered !**

PR #360, opened 2016-01-02
Added automatic model loading from mod jar.

cm0x4D:
So we can get rid of that immense list of strings :-D


cm0x4D:
@Dolu1990, @lambdaShade please review.

cm0x4D:
If someone could test on Windows and Linux, on Mac it works for in Dev environment and in Minecraft itself.

lambdaShade:
Okay for me !

metc:
I will review this soon. I will first merge #353 so assets are clean.
I can test it on Windows and in a few days on Linux if needed.

cm0x4D:
Yes. Works fine on Mac. @metc as soon as I know that the model loading works on Windows too, I will merge. If Mac works, Linux should not be a problem...

src/main/java/mods/eln/misc/Obj3DFolder.java
```diff
@@ -2,59 +2,60 @@
...
-						    }
-						}	
-					}
-				}
-			} catch (URISyntaxException e) {
-				e.printStackTrace();
-			}
-		}*/
+  /** Load all obj models available in the release mod asset folder. */
+  public void loadAllElnModels() {
```
metc, line 21: The debug output is available here: http://pastebin.com/36fPVW47
- Many errors are reported because a lot models do not have txt files
- 101 models are loaded, even if they are not used in the game...

metc:
I have moved the unused models to a separated folder so models which are not yet integrated in the game are not loaded and not included in the release jar file.

This is not the latest build script on this branch, but I have tested with the latest one. Textures files have also been renamed in the master branch, so we will have to test it again when merge.

LGTM, you can merge this PR @cm0x4D.

PR #422, opened 2016-04-16
Syncing up baughn-dev, part 1

Baughn:
These are the _easy_ bits, and the ones I didn't feel I needed to clean up (much).

This PR consists of:
- Adding support for Kotlin.
- Adding a large rheostat, which I programmed in Kotlin, more or less as an example of that.
  We could simplify a lot of the code by converting it, but that'd be pretty invasive. So I haven't done that.
- Makes the autominer a bit nicer for server owners (such as myself), and lets you balance it against the others.
- And fixes two issues that broke the use of IDEA for me.


Baughn:
I'm not sure how to properly translate "Large_Rheostat.name" under the new regime, so an explanation of that would be handy.

cm0x4D:
Thanks for the cleanup and the PR. I will have a look at the Kotlin string translation tonight or tomorrow morning. The changes will then hopefully explain the new translation system.

cm0x4D:
Ok, the translation is trivial, as all is done in the Eln.java file, so Kotlin support is not (yet) needed.

I played with the Rheostat and it is quite nice, but for my taste the model is too similar to the dissipator models, so there might be confusions. Do you mind if I try to modify the model in order to distinguish it from the dissipators?

The pull request LGTM, but we are already pretty late for r51, so maybe we should make the r51 and merge the pull request for r52. What do you think @Electrical-Age/devs?

Baughn:
For what it's worth, I've run this code for months on my own server, and it's been stable. I can't guarantee that no glitches snuck in during cleanup, however.

Personally I'm for cutting the r51 release branch right now, then merging this afterwards. Professional paranoia. :)

Re. model, no, go ahead. I learned Blender entirely so I could work on Eln, and I still kinda suck at it. It makes sense to me that they'd look similar—they do basically the same thing—but, well, I'm not an electrical engineer.

Baughn:
Just to clarify, my suggestion is that we do this:
- Create an "r51" branch from the current 1.7.10-MNA. (And perhaps we should rename that to "master".)
- Continue to develop on master, including merging PRs whenever it seems reasonable; I'd prefer not to have releases block development.
- Cherry-pick individual commits to the release branch if bugs show up, so as to avoid introducing extra bugs from normal development.

Does that seem sensible?

cm0x4D:
Seems reasonable. What do you think about git-flow? This could automate a lot of things and it's basically what @Baughn proposes.

Maybe we could discuss the version number system too. BETA-1.10r51 it is imho not very significative and it can somehow be misunderstood by users. How about a versioning scheme using the year and month of the release (16.04) or just the release number (Release 51)?

Baughn:
My own preference is [semver](http://semver.org/). I've seen it used on several projects, and it generally works well.

Baughn:
Merged as #425.

PR #427, opened 2016-04-23
feature/rheostat wiper

cm0x4D:
@Baughn if you are ok with the model & animation, we can merge this feature into develop...


Dolu1990:
Greate : D

Issue #393, opened 2016-01-26
Increase Batterycapacity


Domi1993:
I just calculated the Capacity of the Batteries (the blockshaped not the Items), and i think the Batteries are (to small), they have unreal small capacity for their size. It is a good investition to increase Their Capacity for the Future is useful when even more Things are coming.

```
                               Old                               New
- Cost Oriented                 60KJ           333.33mAh          1800KJ      10Ah
- Capacity Oriented             240KJ          5.333Ah            4500KJ      50Ah
- Voltage Oriented              60KJ           83.33mAh           3600KJ      5Ah
- Current Oriented              40KJ           222.22mAh          2700KJ      15Ah
- Life Oriented                 60KJ           333.33mAh          3600KJ      20Ah
```

I Just suggest this Values, we can handle it when you think they are too high.

Dolu1990:
Initially i don't have designed these capacity to be realistic, but to create a gameplay. I think new capacity in you proposal are why to high. But i'm not against adding a factor in the configuration file :)

Domi1993:
I agree but the factor is a very good idea so everyone can define his own Capacitys i prefer higher because i play in Creative

Baughn:
FWIW, when I was running a TPPI base powered purely by EA, I found that the battery capacities were small enough to make them difficult to use. As in, I needed a bank of 64 high-capacity batteries to duplicate the functionality of a single EnderIO cap bank.

I did so anyway, because it was fun, but if I hadn't turned off explosions the resource drain from occasional mistakes that make every one of them explode would have been painful. (Which is more of an issue with the explosions. Expect some related PRs in the next two or three weeks.)

On the other hand, with a pure Eln base, that same bank would have been massively excessive. So I'm all for the option, but perhaps document that it should be set to 2 or 4x when used in larger modpacks.

bloxgate:
I'd certainly like to see something like this. For me the Eln batteries have always felt a bit small unless you're using massive banks.

MaHuJa:
ELN cost battery: 60 kJ
Real car battery: 216 to 432 kJ (10 Ah)
@Domi1993's suggestion: 1800 kJ (10 Ah)

I don't know which of these are the result I'm looking for. I would presume the first, the result being about a 'capacitor'.
http://www.wolframalpha.com/input/?i=10+Ampere+hours+at+12V
http://www.wolframalpha.com/input/?i=10+Ampere+hours+at+12V+in+joules

---
[...]

PR #428, opened 2016-04-24
Battery capacity is configurable in Eln.cfg balancing.

cm0x4D:
Battery capacity can be adjusted as factors for generation can be adjusted too.

Added minimal protection for overcharge when factor is reduced. Not a single battery exploded on the tutorial map, so it looks promising. More tests need though...

---

This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/428)

### DIFF
--- a/src/main/java/mods/eln/Eln.java
+++ b/src/main/java/mods/eln/Eln.java
@@ -291,7 +291,8 @@ public class Eln {
 	public boolean killMonstersAroundLamps;
 	public int killMonstersAroundLampsRange;
 
-	double stdHalfLife = 2 * Utils.minecraftDay;
+	double stdBatteryHalfLife = 2 * Utils.minecraftDay;
+	double batteryCapacityFactor = 1.;
 
 	@EventHandler
 	public void preInit(FMLPreInitializationEvent event) {
@@ -373,7 +374,8 @@ public class Eln {
 		Other.ElnToTeConversionRatio = config.get("balancing", "ElnToThermalExpansionConversionRatio", 1.0 / 3.0 * 4).getDouble(1.0 / 3.0 * 4);
 	//	Other.ElnToBuildcraftConversionRatio = config.get("balancing", "ElnToBuildcraftConversionRatio", 1.0 / 3.0 / 5 * 2).getDouble(1.0 / 3.0 / 5 * 2);
 
-		stdHalfLife = config.get("battery", "batteryHalfLife", 2, "How many days it takes for a battery to decay half way").getDouble(2) * Utils.minecraftDay;
+		stdBatteryHalfLife = config.get("battery", "batteryHalfLife", 2, "How many days it takes for a battery to decay half way").getDouble(2) * Utils.minecraftDay;
+		batteryCapacityFactor = config.get("balancing", "batteryCapacityFactor", 1.).getDouble(1.);
 		
 		ComputerProbeEnable = config.get("compatibility", "ComputerProbeEnable", true).getBoolean(true);
 		ElnToOtherEnergyConverterEnable = config.get("compatibility", "ElnToOtherEnergyConverterEnable", true).getBoolean(true);
@@ -1263,7 +1265,7 @@ public class Eln {
 					"BatteryBig", batteryCableDescriptor, 0.5, true, true, voltageFunction, stdU,
 					stdP * 1.2, 0.000, // electricalU,
 										// electricalPMax,electricalDischargeRate
-					stdP, stdDischargeTime, stdEfficiency, stdHalfLife, // electricalStdP,
+					stdP, stdDischargeTime * batteryCapacityFactor, stdEfficiency, stdBatteryHalfLife, // electricalStdP,
 					// electricalStdDischargeTime,
 					// electricalStdEfficiency,
 					// electricalStdHalfLife,
@@ -1283,7 +1285,7 @@ public class Eln {
 					"BatteryBig", batteryCableDescriptor, 0.5, true, true, voltageFunction,
 					stdU / 4, stdP / 2 * 1.2, 0.000, // electricalU,
 														// electricalPMax,electricalDischargeRate
-					stdP / 2, stdDischargeTime * 8, stdEfficiency, stdHalfLife, // electricalStdP,
+					stdP / 2, stdDischargeTime * 8 * batteryCapacityFactor, stdEfficiency, stdBatteryHalfLife, // electricalStdP,
 					// electricalStdDischargeTime,
 					// electricalStdEfficiency,
 					// electricalStdHalfLife,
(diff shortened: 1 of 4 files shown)


Dolu1990:
https://www.google.ch/search?q=atomic+shrum&espv=2&biw=1633&bih=906&source=lnms&tbm=isch&sa=X&ved=0ahUKEwiapYqJoqjMAhUsIcAKHVvdAvsQ_AUIBigB#tbm=isch&q=atomic+mushroom

cm0x4D:
Hehe. Yes, did some tests increasing the factor, which the batteries liked a little less. Is there a problem when the (percental) charge becomes very small or is it the abrupt change of voltage/current?

Dolu1990:
Ahhh yes, it could create a big current rush to capacitors/battery because of battery voltage change (as you said)
Maybe the mode should detect at launch nominalQ changes and then adapt the Q charge level of each batteryElement ?

Baughn:
The better solution might be to store the charge as a fraction of battery capacity, if we want to support this scenario.

Dolu1990:
Yes, it's right, it should not be to hard to change, i will do it.

Domi1993:
is this feature inplemented now?

cm0x4D:
Did you mean that @Dolu1990?

Just made a quick test, seems to work on the tutorial map (Factor: 1 -> 1000 -> 0.01).

As batteries can be slightly overloaded, I put the threshold to detect if the charge has to be converted to 1.1, but it was just a guess. How much can they get overloaded before they explode? We should better take that value...

One problem we could still have is when the factor is changed at the same time the map is opened the first time with the new version of the mod...

Domi1993:
How i can get this build with gitbash on my PC?

Domi1993:
I know how to get the 1.7.10MNA. Which command i must use for get it from the battery capacity factor branch

cm0x4D:
@Domi1993 Well it is still quite experimental and if you go back to r50 you might experience this:

kf0a6vv

So make backups of the maps before doing anything else!

Do this in your cloned repository.

```
# git checkout feature/battery-capacity-factor
```

And then build as usual...

Domi1993:
I make regulary backups of my maps, because i experimented with bigger capacitys too by modifing the eln.java and it happens that i forget to change it after clone the mod. And then my Powerplant says goodbye with a nice Boom :(

cm0x4D:
@Domi1993 Could you test the change? Was it ok?

Domi1993:
I dont had the time to test it but eventually tomorow i can test it. my Batterys have now 180KJ instead of 60KJ of energy stored, when they survive it then it's OK

Domi1993:
when i enter "git  checkout feature/battery-capacity-factor" then build it makes the normal version there is no factor for battery capacity

cm0x4D:
You need to start it at least once in order the config file gets updated.
...and it is in "balancing" section and not in "battery" section.

Domi1993:
The batterys has exploded, with factor 1.0 they stored about 223KJ when i tested it.
we need to store the % value in the maps and not the KJ

cm0x4D:
> we need to store the % value in the maps and not the KJ

Well that is exactly what this branch does. 

I still have to find the right value in order to detect if the value is in J or % when loading the block. Could you describe the problem a little more exact (how many batteries, actual charge, output power,...) or send me your map that causes the troubles?

Domi1993:
There are 20 Life oriented batterys, on my map i got them 240KJ (i modified the eln.java then i builded it) there is s section thats for the batterys, i changed stdDischargeTime to stdDischargeTime \* 4. the same does the battery capacity factor now in the code
(stdDischargeTime \* batteryCapacityFactor).

I builded this branch, and then set in eln.config batterycapacityfactor to 1.0

Every of the 20 Batterys have power from -2W at night ... 7W at day
I think the current was not the problem, they blowed up just when i opened the map!

While The ELN Blocks are loading it happened

cm0x4D:
It is very important that the map is opened at least once with the same factor as the values of the batteries have to be converted from J into % of maximal charge. So if you load your map for the first time with this branch, you need to set the factor to the same value as you had in the code (I think it would be 4). Otherwise the abrupt change of charge will produce current peaks that can destroy not only the batteries...

Domi1993:
I'll try it when i have time. That means the mod will read first time the J of my Batterys and store it as %. When i do the factor to 8 my batterys are charged at 100% right? But is that not a little cheaty when i can set a low factor to charge it and then put it to i.e. 20.0 and use energy i don't produced? 

I test out this what happens and tell it here

cm0x4D:
> But is that not a little cheaty when i can set a low factor to charge it and then put it to i.e. 20.0 and use energy i don't produced?

Yes that's in fact the case. But if we change the battery charge (%) at startup to much, they will explode like yours did, so we do not really have an option here.

Baughn:
Someone who can change the server configuration could also, for example, set themselves to creative and spawn in a voltage source. I can think of no serious reason to worry about cheating here.

Domi1993:
Yes but i just think when someone did this in single player

cm0x4D:
It is open source, we can not avoid cheating anyway, as anyone can change that factor in Eln.java already like you actually did. Basically as soon as you touch Eln.cfg you are actually already some kind of cheating...

Baughn:
If you're in single player, you can always set yourself to creative. And if you're playing Eln with some kind of modpack, then there will always be other mods that are more powerful or at least easier to use, so...

I think we need to trust the players to not ruin their own fun. :)

Domi1993:
It's working now my Batterys have survived with 240KJ to 30KJ.
it's now the percent thats stored in the map

cm0x4D:
Good to hear. I guess some more tests are still needed...

Domi1993:
I think we need a lower limit for the Batterycapacityfactor.

I can test out which factor will be dangerous (low capacity) if you want

cm0x4D:
Ok. Some tests would be nice, but imho if someone puts in there ridiculous values he will get what he deserves...

Domi1993:
Is this feature avialable in the new release?

metc:
@Domi1993 No, these changes are not yet merge in the develop branch.

FYI, all changes, fixes and improvement of the r51 release are described here: https://electrical-age.net/news/release-1-11/

metc:
Related to #393

cm0x4D:
I need to ensure that the batteries are not empty when converting to per-unit. The decision is actually made using the Q value: If it is smaller than 1.1 we assume that Q is saved as per-unit, all values bigger than 1.1 will be converted to per-unit. This can lead to problems for very empty batteries - They will instantly go from almost 0% to 110% of charge. This does not end well!

I have to do some additional check in order to avoid the case above - This should be quite simple by checking the actual voltage of the battery according to the nominal voltage.

After that the feature can be merge and be part of the next release...

cm0x4D:
Reviewed 1 of 2 files at r1, 2 of 3 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/electrical-age/electricalage/428)_

Issue #269, opened 2015-07-18
[Question][Mod Interaction] Power Conversion


ReikaKalseki:
Is there a system by which I could add converters to and from ElectricalAge/RotaryCraft power? If so, how, and are there balance or progression factors I need to take into account?

Issue #215, opened 2015-04-12
Request: RotaryCraft support!


BionicFelps:
RotaryCraft seems to have a lot of points for mod interaction with this mod.  It has a two-factor power system (torque, speed), it has a concept of heat as well, it also has an API for easy integration.  I'd recommend allowing this mod's heat items to operate on RotaryCraft heat (heat sinks, heat furnace, etc.) and also creating a rotational power -> ELN converter (generator).  Power can already go the other direction via the ELN->Other mod block and an RF powered RotaryCraft generator.

I'm currently making a Modpack that will include both mods and see a lot of potential here.  The pack will NOT include Thermal Expansion, BuildCraft, or any of the other "traditional" power mods, as it's designed to highlight a number of less-known mods, and I believe I found a nice second ecosystem of mods that should surprise people!

RotaryCraft also has an addon called "ReactorCraft" that adds complex nuclear reactors, which take a lot of resources and know-how to construct but produce vast amounts of rotational power.

BionicFelps:
I'm assuming that you would have made other to ELN blocks with the ELN to other blocks if you wanted RF to be used for power, assumption being that RF is not two-factor power, so you would just have to automatically make it at 50v, 200v, and 800v and a variable amps depending... Also energy cells would trivialize the use of batteries.  

RotaryCraft has no easy power storage itself, along with the two-factor power system, so it seems it would take care of those concerns automatically!  In fact, ELN batteries would be a nice way to store rotarycraft power at that point...

nanakisan:
Hmmmm ELN's power system is about as equally complex as RotaryCrafts. However ultimately this would require Dolu having Reikas permission. Reika is very and I mean very strict on compatibility of mods with his. If anything Reika might add the compatibility himself instead of allowing Dolu to add a block for it.

BionicFelps:
This is true, he's also quite strict on modpack devs, I didn't want to bring it to Reika himself because he actually makes an electrical mod with full RotaryCraft support.  I mistakenly thought that his API was open for mod interaction though.  (As essentially the writer of the API is still in control of what's possible, I didn't check..)

ghost:
I'm aware this is a bit late, but Reika has mentioned that he's not against adding converters to his mods, as long as he can talk to Dolu first.  ReikaKalseki/Reika_Mods_Issues#109

Personally, once that hurdle's been passed, I'd like to be able to add this to the Survival Industry modpack.

PR #430, opened 2016-04-26
Bump forge, add mechanical shafts, steam turbine and generator

Baughn:
The forge bump is necessary because there was a breaking API change to fluid handling in-between. So I bumped it to current stable, but we could go a few versions back if you want.

Otherwise, this code is pretty thoroughly tested. We've been running it on Erisia for the last couple of months, which gave me a chance to balance it as well.

The steam turbine could be argued to be an other-to-eln energy converter. I did include some interesting game mechanics:
- The spinning shafts have a lot of inertia, and the generator can work as a motor, albeit at 75% efficiency. This results in the generator acting as a sort of capacitor; just like real-life mechanical generators, we've found that they tend to stabilize large-scale power grids against sudden spikes or dips.
- The efficiency at which the steam turbine converts steam to shaft power drops dramatically at low speeds. (And at overspeed, but that's not usually relevant.) This means you need prodigious amounts of steam to start it, _or_ you can spin it up with a generator operating in reverse. Preferably without explosions. Rheostats help for that.
- The steam turbine, incidentally, uses an exponential decay function for steam input. This means it's slow to stop; loss of load will usually cause it to explode, so you need an emergency dump. Not completely realistic, but neither is steam in minecraft overall. I have plans to improve this. (A 'steam injector' block that allows you to dump it, but not just suddenly stop.)
- There's about 400W of drag in a generator/turbine combo at nominal speed. They're not efficient for light use, but you can get 12kW out of three generators plus a turbine, so it's <10% when used optimally.

The upside, of course, is that it's relatively easy to generate vast amounts of power using this. Railcraft boilers are far more efficient than heat turbines, and of course the various reactor mods tend to generate steam.

In practice, players tend to use it for base-load.

The TransparentNode subclass code should also make it feasible to finally finish the Fuel Generator. (Who was working on that?)


Baughn:
I also intend to do a bit of cleanup on the generator and turbine code, specifically "SimpleShaftElement" et al, but I'd very much rather have this merged so I can do that without needing to painfully rebase _again_. I hope that's okay. :X

cm0x4D:
There is a [request](https://github.com/Electrical-Age/ElectricalAge/issues/215) for adding RotaryCraft support, the generator could either serve as base for a new block that offers such an integration or maybe RotaryCraft support could be added to the generator itself.

@Electrical-Age/devs What is your opinion?

@Baughn The generator produces iirc 3kV, so according to the color scheme that simplifies the usage of the mod, we should change the color to a kind of red. Do you mind if I change the color on the texture to a red?

Baughn:
Re. Rotarycraft: Can't be done without serious thought. There are feature, scale and balance mismatches. Off the top of my head:
- RotaryCraft engines run at fixed RPM and torque. If that suffices for the powered machines, they run; otherwise, not. My code simulates shaft inertia instead, so there's the question of what to do when RPM is mismatched. Even in a simple turbine/generator pair, RPM will fluctuate with voltage, and there's no simple notion of 'torque' that can be read out. I'd like for any integration to be two-way, subject to possible balance issues on the RotC side.
- RotaryCraft operates on a kind of log-base2 power system. (Ignoring Electricraft) all engines produce 2^N kW, from 16kW for the steam engine to multiple GW for large reactors. To balance this out, the operation speed of its machines typically take the logarithm of input speeds (or power, or min(speed,power)) again. He has reasons for this, and it works well internally to the mod, but a conversion block will need to do something similar. I'd like to hear @ReikaKalseki's thoughts on balancing this.
- Linear scaling. I've thought this before, but... that 16kW steam engine is the second-weakest engine that's possible to build, and we can't very well have it produce 16kW on Eln's scale. See above, though.

That being said, RotaryCraft is one of my favorite mods, as well as the explicit inspiration for this bit of code. I'd love to have integration.

Re. colors: I wouldn't mind. I may be ok-ish at making models, but I suck at painting them. I don't think this should block a merge, but you might put it as a blocker for the next release.

We should also think about what color to use for the 800V tier, which is probably where I'm going to stick the chunkloader.

ReikaKalseki:
Though you can easily make a block that converts from RC to ELA/ELN power, the reverse is impossible so as to prevent mods from unbalancing the RC progression. What that then means is that if you want design symmetry the converters would probably have to be added on my end, something which [I actually asked about a year ago and was closed recently](https://github.com/Electrical-Age/ElectricalAge/issues/269). However, such generators operate "simply", like a magnetostatic, taking and accumulating energy for the to-RC conversion and simply constantly generating some value (based on input power) for the from-RC conversion. There is no capability for more complex behavior like EMF feedback. Also, I do not have a formal education in electrical engineering outside of two undergraduate courses, so I do not have the same theoretical background I do for mechanical (or, less relevant, thermodynamic or aerodynamic) systems.

Things like heat, however, as per [the initial linked request](https://github.com/Electrical-Age/ElectricalAge/issues/215), are indeed fully compatible; simply implement the relevant interfaces and the Friction heater or similar will work on your machines (using my temperature scales, but I suppose a multiplier parameter is easily doable, and will implement that later today).

As for scaling differences between the power...I am not sure. I would be hesitant to have the converters apply a nonlinear operator like a log during the conversion, as from what I understand ELN has RF or other power converters, and the other RC converters would remain linear. Thus if, say, ELN was log2(RC) in the converter, you could bypass that by converting it through RF first. I also do not want to have people start asking for the other conversions to be similarly throttled, as I very much like the massive RF capabilities of endgame RC (and it in fact was one reason I changed the conversion rate in v4 to be 9x more RF/watt).

Baughn:
The lack of feedback is basically fine. Right now, offering an API is unfortunately difficult, but I'll keep it in mind for later...

Re. RF-to-Eln conversion: This doesn't exist. The conversion blocks that exist today are solely Eln-to-other, with this particular PR adding steam-to-Eln. That's for... pretty much the same reason as yours, actually--balance.

But this PR is the wrong place to discuss it, and some of the design decisions I'm basing my statements on exist only in my head. I'll get back to you later, once things are a bit more settled.

cm0x4D:
> We should also think about what color to use for the 800V tier, which is probably where I'm going to stick the chunkloader.

Voltage Colors:





It is actually not mandatory but might help the users...

Baughn:
Somehow, I missed that it matches the cables. Whoops. ^^;;

PR #431, opened 2016-04-30
Feature/signalprocessor

Baughn:
Just some simple enhancements while I work on the powerline code.


cm0x4D:
Adding pow() as function is a imho good idea. Removing the exponential operator might cause problems on existing maps, so I would prefer to offer both, the function and the operator. Otherwise LGTM

Baughn:
This isn't removing the exponential operator, though. It's removing a commented-out reference to it.

cm0x4D:
Duh! Well then lets merge ;-)

PR #432, opened 2016-04-30
AutoMiner: Add silk touch option

Baughn:
What it says.

At some future point I'd like to be a bit more clever about this, but this should do for now.

### DIFF
--- a/src/main/java/mods/eln/gui/GuiContainerEln.java
+++ b/src/main/java/mods/eln/gui/GuiContainerEln.java
@@ -56,7 +56,7 @@ public abstract class GuiContainerEln extends GuiContainer implements IGuiObject
 	}
 
 	public GuiButtonEln newGuiButton(int x, int y, int width, String name) {
-		GuiButtonEln o =  helper.newGuiButton(x, y, width,name);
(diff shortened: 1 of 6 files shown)


Baughn:
The constant indentation confusion is getting a little confusing. At the very least, it confuses IDEA. Everything ends up looking right, to _me_, but then I get diffs like this...

It really wouldn't be a bad idea to reformat the entire project, I think.

Baughn:
For amusement value, this is what that diff looks like: https://github.com/Electrical-Age/ElectricalAge/commit/22640288a050b571e7408e631a4d661930e5828b

cm0x4D:
LGTM.

What we could do when creating GUI's is to take all available space for buttons and all elements that could contain text. So the chances that localized strings does not overflow would be minimized.

I can do it after the merge or you can do it before, like you prefer...

Baughn:
I cargo-culted this UI into existence. I honestly don't understand the UI code at all, so I'll leave that to you, I think.

cm0x4D:
Ok. Will try to change that. I agree that to code UI's is far away from a joy ;-)

PR #435, opened 2016-05-04
Fuel generator 50V/200V

cm0x4D:


src/main/java/mods/eln/transparentnode/FuelGenerator.kt
```diff
@@ -0,0 +1,291 @@
...
+
+class FuelGeneratorElement(transparentNode: TransparentNode, descriptor: TransparentNodeDescriptor):
+        TransparentNodeElement(transparentNode, descriptor) {
+    internal var positiveLoad = NbtElectricalLoad("positiveLoad")
+    internal var powerSource = PowerSource("powerSource", positiveLoad)
+    internal var slowProcess = FuelGeneratorSlowProcess(this)
+    internal var descriptor = descriptor as FuelGeneratorDescriptor
+    internal val fuel = FluidRegistry.getFluid("fuel") ?: FluidRegistry.getFluid("lava")
+    internal var tankLevel = 0.0
+    internal var on = false
```
Baughn: This could usefully be done with a delegate, aka. https://kotlinlang.org/docs/reference/delegated-properties.html

src/main/java/mods/eln/transparentnode/FuelGenerator.kt
```diff
@@ -0,0 +1,211 @@
...
+import org.lwjgl.opengl.GL11
+import java.io.DataInputStream
+import java.io.DataOutputStream
+
+class FuelGeneratorDescriptor(name: String, internal val obj: Obj3D?, internal val cable: ElectricalCableDescriptor,
+                              internal val nominalPower: Double, internal val maxVoltage: Double,
+                              tankCapacityInLiters: Double)
+: TransparentNodeDescriptor(name, FuelGeneratorElement::class.java, FuelGeneratorRender::class.java) {
+    companion object {
+        val FuelCalororificValue = 30500.0;                         // Fuel has a calorific value of about 30.5 MJ/l
```
Baughn: "Caloric value"

Baughn: "Calorific" - varname misspelled.

src/main/java/mods/eln/transparentnode/FuelGenerator.kt
```diff
@@ -0,0 +1,211 @@
...
+    override fun shouldUseRenderHelper(
+            type: IItemRenderer.ItemRenderType, item: ItemStack,
+            helper: IItemRenderer.ItemRendererHelper) = type != IItemRenderer.ItemRenderType.INVENTORY
+
+    override fun renderItem(type: IItemRenderer.ItemRenderType, item: ItemStack, vararg data: Any) {
+        if (type == IItemRenderer.ItemRenderType.INVENTORY) {
+            super.renderItem(type, item, *data)
+        } else {
+            objItemScale(obj)
+            GL11.glPushMatrix()
```
Baughn, line 91: Consider using the 'preserveMatrix' utility I added: https://github.com/Electrical-Age/ElectricalAge/blob/develop/src/main/java/mods/eln/misc/GLUtil.kt#L10

src/main/java/mods/eln/transparentnode/FuelGenerator.kt
```diff
@@ -0,0 +1,211 @@
...
+        this.descriptor = descriptor as FuelGeneratorDescriptor
+    }
+
+    override fun draw() {
+        renderPreProcess = drawCable(Direction.YN, descriptor.cableRenderDescriptor, eConn, renderPreProcess)
+        front.glRotateZnRef()
+        descriptor.draw()
+    }
+
+    override fun cameraDrawOptimisation(): Boolean {
```
Baughn: override fun cameraDrawOptimisation() = false

src/main/java/mods/eln/transparentnode/FuelGenerator.kt
cm0x4D: Good idea. I will do that. Thanks!

cm0x4D: I added a custom property delegate that can be used to automatically publish properties when set: https://github.com/Electrical-Age/ElectricalAge/blob/feature/fuel-generator/src/main/java/mods/eln/node/Publishable.kt

cm0x4D:
Ow. The recipes are still missing. As I am rather the creative player, does someone has any idea or suggestion for a recipe? Thanks!

metc:
This new item looks great Michi ;) More than 300 lines of noisy Java replaced with Kotlin, this is interesting. The two recipes you choose make sense for me.

Baughn:
I would say, add a a cheap/advanced chip respectively to balance it out a little, and to force people to touch other forms of power generation as well.

src/main/java/mods/eln/transparentnode/FuelGenerator.kt
```diff
@@ -0,0 +1,287 @@
...
+
+    override fun shouldUseRenderHelper(
+            type: IItemRenderer.ItemRenderType, item: ItemStack,
+            helper: IItemRenderer.ItemRendererHelper) = type != IItemRenderer.ItemRenderType.INVENTORY
+
+    override fun renderItem(type: IItemRenderer.ItemRenderType, item: ItemStack, vararg data: Any) = when(type) {
+        IItemRenderer.ItemRenderType.INVENTORY -> super.renderItem(type, item, *data)
+        else -> {
+            objItemScale(obj)
+            GL11.glPushMatrix()
```
Baughn, line 91: Doing this with preserveMatrix is better - rather less chance of error.

cm0x4D, line 91: Agree, but I do not have this function yet in my branch, so I prefer to change that after the merge (I know I am lazy ;-))

Baughn, line 91: Fair 'nuff.

Baughn:
Can cheap chips be made without power? I don't recall, off-hand. If they can, then...

I don't suppose there's a huge risk of players sticking to this generator and never building anything else. I'm probably just being a scaredy-cat.

cm0x4D:
The sound of them is pretty annoying and they last at least 20 minutes at full load, so I do not think that they will be part of permanent installations. 

They actually do not have a GUI. I asked myself multiple times if I should add a GUI. Probably it would be better with GUI, but GUI programming is really a pita, so I choose the short path...

Could be cheap chip or even an analog regulator, as there is really some regulation going on on real generators.

Baughn:
Using an analog regulator makes sense to me.

EDIT: Oh, and for my own code, I prefer to have as few GUIs and as much in-world interaction as possible. Any information that needs to be displayed can be displayed on the model, etc... that was the motivation behind the LEDs on the generator.

PR #439, opened 2016-05-09
Fixes #279

### LINKED ISSUE #279: [Feature-Request] Wireless TX Range config
Increasing/decreasing the wireless range for transceivers and receivers would be great.

cm0x4D:
Allows to configure the wireless TX range in Eln.cfg

### DIFF
--- a/src/main/java/mods/eln/Eln.java
+++ b/src/main/java/mods/eln/Eln.java
@@ -424,6 +424,8 @@ public class Eln {
 		thermalFrequency = config.get("simulation", "thermalFrequency", 400).getDouble(400);
 		cableRsFactor = config.get("simulation", "cableRsFactor", 1.0).getDouble(1.0);
 
+		wirelessTxRange = config.get("wireless", "txRange", 32).getInt();
+
 		config.save();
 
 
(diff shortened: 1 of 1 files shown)


cm0x4D:
@Electrical-Age/devs: Any opposition against this feature? Can I merge?

Baughn:
In case it becomes relevant: I'm planning to add a powered-antenna multiblock at some stage. I'm not opposed to this one, especially as I have no idea when that might happen.

metc:
I did not play with this feature, but short question: what is appending if the TX range is a negative value ? Should you not check the bounds of the user input ?

cm0x4D:
I guess a negative value would be something like the flight mode on your mobile ;-)

cm0x4D:
If someone puts stupid values in the configuration, he should imho get what he deserves...

metc:
LGTM, you can merge.

General remark: each descriptors should maybe have up/down limits for each values so we could check the bounds automatically and print warning messages when required.

cm0x4D:
Agree, would be good for all settings. I will create an issue for that...

Baughn:
Forge's configuration API already comes with support for range limits. You should use that, if possible...

PR #440, opened 2016-05-11
Feature/sixnode placing direction support

cm0x4D:
Not quite ready to merge, but I would like to start the discussion about this feature.

I would like to achieve that each SixNodeDescriptor can determine the front direction of the element at the creation by overriding a method like it is possible for transparent nodes (getFrontFromPlace() method).

I did the needed changes in SixNode code to enable this and changed for all six node elements that did this in the constructor to use the actual method in the descriptor.

In the data logger I experimented with the idea, that if you place a block on a wall, the direction is always "up", but if you place the block on the ground or the ceiling, up is in the direction the player is actually looking horizontally. This would improve the user experience as most of the time six nodes are placed automatically as the user expects it.

@Electrical-Age/devs What do you think?

### DIFF
--- a/src/main/java/mods/eln/Eln.java
+++ b/src/main/java/mods/eln/Eln.java
@@ -2348,7 +2348,7 @@ public class Eln {
 			name = TR_NAME(Type.NONE, "High Voltage Relay");
 
 			desc = new ElectricalRelayDescriptor(
-					name, obj.getObj("relay800"),
+					name, obj.getObj("relayBig"),
 					highVoltageCableDescriptor);
 
 			sixNodeItem.addDescriptor(subId + (id << 6), desc);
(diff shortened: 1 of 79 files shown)


Baughn:
Haven't looked very closely at the code yet.

I like the UI side of this. It isn't a massive improvement, but small quality-of-life improvements are always welcome. Code-wise... you're returning LRDU.Down for most of these, so it might make sense to have that as the default instead of overriding in every single SixNode.

metc:
This is a great feature. Agree with @Baughn IMO, all blocks should have the same "default" behavior. This new one makes sense.

cm0x4D:
Valid point @Baughn. I did actually not count the nodes that changed that from Up in the constructor. I saw in the constructor it is LRDU.Up, so I used that in order to not change their behavior and moved the modification that was present to the new appropriate method.

That is only an intermediate state, finally I would like change the SixNodeDescriptor implementation to something similar as now the data logger, so this behavior would be standard for all elements.

That said I discovered that it might be actually more complicated as I thought, as for some elements only the connection position is changed, but the model is always rendered in the same direction. I guess this behavior is good for walls, but not necessarily on the floor...

So I have to still some work to do on this...

cm0x4D:
If I have not forgotten any sixnode, it should be done now.
Feel free to test or review. I would say there is no problem merging it now.

PR #443, opened 2016-05-16
Feature/build in modbus

cm0x4D:
The modbus4j dependency for just the Modbus RTU which is not used by many people has been removed in this branch and replaced by a very basic Modbus TCP Slave implementation.


Baughn:
I don't even know what it's supposed to do. :P

That's probably why not many people use it.

cm0x4D:
Using this you could theoretically automate your Minecraft world using a real world PLC. We use this in a practical work at our university to teach SCADA principles.

metc:
LGTM and makes sense.
This is probably the first Slave Modbus RTU implementation written in Kotlin <3

cm0x4D:
Even if it would not work perfect yet, it is definitely an improvement as the jar looses about 500 kB of weight and the feature is used by nobody anyway...

PR #445, opened 2016-05-25
Fixed rotation direction in the case shaft elements are placed in opp…

cm0x4D:
…osite direction


cm0x4D:
@Baughn Playing around with the mechanical shaft elements I discovered if they are placed in opposite direction, that the shaft rotation animation was in opposite direction too. 

I fixed this be defining a rotation direction for positive X and Z directions and the opposite rotation direction for negative X and Z. You can merge it into develop if you are ok with the changes...

Baughn:
Fixing that glitch was sitting deep on my todo-list, so I certainly have no objections. Thanks. :)

PR #447, opened 2016-05-30
Update contributions

Baughn:


cm0x4D:
@Baughn imho the main developer section would better suit your actual status.

metc:
Should we also add a link in the language section with all the contributors who help to translate the mod on https://www.transifex.com/electrical-age/teams/ ? Whould be great imo

PR #453, opened 2016-06-04
Fix Modbus TCP Server

### LINKED ISSUE #451: Issue with ModBus
Running this concurrent on the same box produces an error even if modub is not enabled.  Disabling the mod allows the server to boot.

See end of crashlog:

http://pastebin.com/404Df3Sh

Thanks for your time.

bloxgate:
Closes #451 
The server was attempting to bind to port before checking if it was enabled or not.
This PR:
- Only attempts to bind if enabled, closes server otherwise.
- Makes the port configurable
- Catches exceptions, and let's the user know there's a problem instead of crashing if bind fails.
- Adds some directory IntelliJ spat out to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ eclipse
 *.json
 options.txt
 /out
+/classes
 
 # IntelliJ
 .idea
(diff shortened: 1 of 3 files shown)


cm0x4D:
Big thanks!

PR #454, opened 2016-06-05
Feature/better turret

Baughn:
In short, this patch allows players to use turrets for e.g. wither farms.

I'll admit that the reason I wrote this is that... well... I already made the farm. And a massive power system. And a gas turbine to drive the power system. And architecture. And I didn't feel like swapping to a killer joe. :P


Baughn:
Here's how it looks: http://i.imgur.com/a5UHySw.jpg

Baughn:
Withdrawn for the moment. There seems to be a bug or two.

PR #459, opened 2016-06-24
Feature/gas turbine

Baughn:
Don't merge _yet_, but let's start discussing this. Things I believe still need doing:
- [x] Some of these commits need better descriptions.
- [x] I think, on balance, the steam turbine shouldn't have a throttle. Or it should drop the excess steam on the floor.
- [x] I have a long list of suitable fuels. This should be factored out, and the Fuel Generator should use the same / a similar list. Not everyone runs Buildcraft.
- [x] Still need a better gas turbine model. Something that looks more like http://cdn.wartsila.com/images/default-source/Power-Plants-pictures/gas-turbine-for-power-1.jpg?sfvrsn=2


cm0x4D:
> I have a long list of suitable fuels. This should be factored out, and the Fuel Generator should use the same / a similar list. Not everyone runs Buildcraft.

Agree. We could extend the FluidRegistry by a method to ask if a given fluid can be used as burnable fuel and maybe even attach the energy density of the fuel...

Baughn:
Somewhat of a model: http://i.imgur.com/Jyvj3KV.jpg

I certainly won't mind if one of you improves/replaces it, but this'll do for now.

Baughn:
@cm0x4D: I think this is about complete. Mind taking a look?

src/main/java/mods/eln/Eln.java
```diff
@@ -5241,6 +5251,16 @@ void recipeTurbine() {
 				Character.valueOf('A'), "blockAluminum",
 				Character.valueOf('M'), findItemStack("Advanced Machine Block")
 		);
+		addRecipe(findItemStack("Gas Turbine"),
+				"msH",
+				"sSs",
+				" M ",
+				Character.valueOf('m'), findItemStack("Advanced Electrical Motor"),
+				Character.valueOf('H'), findItemStack("Copper Thermal Cable"),
+				Character.valueOf('s'), "ingotSteel",
```
cm0x4D: afaik ingotSteel does not exists in Vanilla Minecraft + EA. Imho recipes should only contain Minecraft stock items and EA items.

cm0x4D:
Lgtm, I propose we merge first your changes and then I rebase and modify my branch, as there are some substantial changes in the Shaft base classes.

src/main/java/mods/eln/Eln.java
Baughn: One thing I've done previously is to use other-mod ingots if they exist, and vanilla/EA otherwise. I'll do that here as well.

It's all temporary until I get the chemical simulation running, anyway. Although I think I want to avoid writing the minecraft interface for that until we're on 1.10.

PR #467, opened 2016-06-29
Mechanical joints

cm0x4D:
Basically two types of mechanical connection joints.

The straight version just connects two shafts from on side to the other.
The Shaft hub connects shafts from all 4 sides.

Both do not store that much energy. I plan adding a flywheel later for this.

Another idea is to add a new sensor that delivers a signal output voltage that is proportional to the rotation speed of the shaft. This would enable to regulate the rotation speed using for example a rheostat.

@Baughn:
- I set a very small mass for these shafts, I really would like to hear your opinion to these in order to balance it with the other mechanical elements.
- I had to make changes to the ShaftNetwork in order to notify ShaftElements about their connection state. Please review the code and ensure that I did not broke anything.


src/main/java/mods/eln/mechanical/Flywheel.kt
```diff
@@ -0,0 +1,17 @@
+package mods.eln.mechanical
+
+import mods.eln.misc.Obj3D
+import mods.eln.node.transparent.EntityMetaTag
+import mods.eln.node.transparent.TransparentNode
+import mods.eln.node.transparent.TransparentNodeDescriptor
+
+class FlywheelDescriptor(baseName : String, obj : Obj3D): SimpleShaftDescriptor(baseName,
+        FlyWheelElement::class, ShaftRender::class, EntityMetaTag.Fluid) {
```
Baughn: Why Fluid?

src/main/java/mods/eln/mechanical/ShaftNetwork.kt
```diff
@@ -78,8 +78,12 @@ class ShaftNetwork(): INBTTReady {
         assert(from in elements)
         val neighbours = getNeighbours(from)
         for (neighbour in neighbours) {
-            if (neighbour.shaft != this) {
-                mergeShafts(neighbour.shaft)
+            if (neighbour.elemment.shaft != this) {
```
Baughn: Misspelled; should be "element".

src/main/java/mods/eln/misc/DirectionSet.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import net.minecraft.nbt.NBTTagCompound
+import java.io.DataInputStream
+import java.io.DataOutputStream
+import java.util.*
+
+class DirectionSet: TreeSet<Direction>() {
+    fun toInt(): Int {
+        var value = 0
+        if (contains(Direction.XN)) value += 1
```
Baughn: This might be clearer if you use |= and (2 << 0), (2 << 1), etc.

Baughn:
This all seems fairly straightforward. Some nits, but no real objections.

One point: I've been meaning to factor out parts of the connection logic, probably to a library which _someone else_ has written. JGraphT, for instance. That has no bearing on this PR, save that you already did some of the necessary cleanup for me.

The mass of the existing blocks were basically pulled out of my ass. It'd be better to look up the mass of actual shafts of this sort, and fix that. Gameplay-wise, it'd be best if they're lighter, so as to provide a role for the flywheel.

> Another idea is to add a new sensor that delivers a signal output voltage that is proportional to the rotation speed of the shaft. This would enable to regulate the rotation speed using for example a rheostat.

So, a dynamometer block? :)

I've said it before, but we can steal a lot from RotaryCraft here.

src/main/java/mods/eln/Eln.java
```diff
@@ -2606,6 +2605,33 @@ void registerTurbine(int id) {
 			);
 			transparentNodeItem.addDescriptor(subId + (id << 6), desc);
 		}
+
+		{
+			subId = 11;
```
Baughn, line 2610: Do you think you could start at 12, reserving 11 for the gas turbine?

Just because I already have it in a large world. I'll try to finish it this weekend.

src/main/java/mods/eln/misc/DirectionSet.kt
cm0x4D: Kotlin does not have |= and << operators afaik.

I would have to do this: 
value = value or 1.shl(x)

Not very readable imho. Maybe I find something smarter than that...

Baughn: It'd be "value = value or (1 shl x)". Not ideal, but not quite that bad either.

cm0x4D: I came up with something like this:

fun Int.setBit(position: Int): Int = this or (1 shl position)
fun Int.isBitSet(position: Int): Boolean = this and (1 shl position) != 0

Do you know if it would be possible to change the value of "this" in an extension function?

src/main/java/mods/eln/Eln.java
cm0x4D, line 2610: No problem.

Baughn:
And here's how it looks: https://brage.info/e12/#17

You're good at modelling. :)

cm0x4D:
The PR is feature complete for:
- Straight Joint
- Joint Hub
- Flywheel
- Tachometer
  So for my part it can be merged into dev now.

Any objections? Any ideas for recipes?

Baughn:
Recipe-wise...

For the straight joint and hubs,

```
...
iii
.m.

.i.
iii
.m.
```

With 'i' being iron ingots, m being a machine block. Cheap, yes, but these aren't functional blocks; I expect to see them used mostly for aesthetic reasons. I view the current recipes as being temporary, though; we should probably have a few more items.

cm0x4D:
@Baughn I have to fix the issue that I actually use ShaftRender directly in my classes. There are basically two options:
- I add some default behavior to the shaft render itself and make it open in place of abstract
- I create some render classes for my items that effectively do not add functionality, but define the abstract attribute to null

I actually prefer the first option.

For the recipes I completely agree.

Baughn:
I'm not sure I quite understand what you're aiming for, but... sure, option 1 sounds better on the face of it.

cm0x4D:
> I view the current recipes as being temporary, though; we should probably have a few more items.
> Agree, but we should have them ready before the next release. I can think of a shaft, a mechanical stand and some cast iron to construct the cases...

@Baughn: I rebased my branch on your  changes in develop and I had a lot to merge manually in Eln.java, can you check that merging this branch will not cause some defects on your side? If not I would merge that branch into develop today.

src/main/java/mods/eln/Eln.java
```diff
@@ -5275,6 +5310,40 @@ void recipeTurbine() {
...
+		addRecipe(findItemStack("Joint hub"),
+			" i ",
+			"iii",
+			" m ",
+			Character.valueOf('i'), "ingotIron",
+			Character.valueOf('m'), findItemStack("Machine Block")
+		);
+
+		addRecipe(findItemStack("Flywheel"),
+			"iii",
```
Baughn: Perhaps use blocks of iron, at the cardinal points?

Baughn:
I don't see any problems in Eln.java.

src/main/java/mods/eln/Eln.java
cm0x4D: You mean:

```
iIi
ImI
iIi
```

Where I is the block of iron?

Baughn: Right.

PR #473, opened 2016-07-06
Update recipes to use ore dictionary names, providing more mod compatibility.

grundyboy34:
Updated recipes to use ore dictionary names to provide better compatibility with other mods.

Also added some logic in the findItemStack method that will search the
ItemRegistry as a last resort, especially when passed an ore dictionary
string, and will return null if it fails in order to prevent
nullpointerexceptions.
- [x] Update recipes to use ore dictionary strings rather than static item/block references.
- [x] Add redundancy checks into findItemStack
- [x] Update addRecipe function to support ore dictionary strings.
- [x] Update scanner to add all ore dictionary ores, fix collisions with metadata items/blocks
- [ ] Create constants for all ore dictionary strings.
- [ ] Create routine to check if ore dictionary strings exist, and add them if not. (Redundancy check)

---

This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/473)

### DIFF
--- a/src/main/java/mods/eln/Eln.java
+++ b/src/main/java/mods/eln/Eln.java
@@ -8,6 +8,7 @@ import cpw.mods.fml.common.network.FMLEventChannel;
 import cpw.mods.fml.common.network.NetworkRegistry;
 import cpw.mods.fml.common.network.simpleimpl.SimpleNetworkWrapper;
 import cpw.mods.fml.common.registry.EntityRegistry;
+import cpw.mods.fml.common.registry.GameData;
 import cpw.mods.fml.common.registry.GameRegistry;
 import cpw.mods.fml.relauncher.Side;
 import mods.eln.achievepackets.AchievePacket;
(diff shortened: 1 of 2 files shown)


src/main/java/mods/eln/Eln.java
```diff
@@ -4891,63 +4892,63 @@ void recipeLampSocket() {
 				"G ",
 				"IG",
 				"G ",
-				Character.valueOf('G'), new ItemStack(Blocks.glass_pane),
-				Character.valueOf('I'), new ItemStack(Items.iron_ingot));
+				Character.valueOf('G'), "paneGlass",
```
Baughn, line 4895: I believe this should be paneGlassColorless

Baughn:
So the basic problem with this sort of change is that the oredict entries may or may not actually exist in base Forge.

One obvious fix is to add the mappings ourselves. Anyhow, it's something you need to check.

src/main/java/mods/eln/Eln.java
grundyboy34, line 4895: Yes it should be, thanks!

grundyboy34:
Forge adds entries like ingotIron and dustRedstone which is where I got those from. I'm going to do another pass through the recipes though to double check everything as well as recipes that I may have put in incorrectly like the paneGlassColorless.

Baughn:
@grundyboy34 -- Did this come to anything?

Baughn:
Closing due to inactivity. Reopen if desired.

PR #478, opened 2016-07-07
Feature/minetweaker

Baughn:
This is just @marcin212 's work, merged into the current version. It'll take a day or two to confirm that it works, then I'll merge it.

---

This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/478)

### DIFF
--- a/src/main/java/mods/eln/integration/minetweaker/utils/RemoveRecipe.java
+++ b/src/main/java/mods/eln/integration/minetweaker/utils/RemoveRecipe.java
@@ -1,0 +1,36 @@
+package mods.eln.integration.minetweaker.utils;
+
+import java.util.ArrayList;
+
+import minetweaker.api.item.IIngredient;
+import minetweaker.api.minecraft.MineTweakerMC;
+import mods.eln.Eln;
+import mods.eln.misc.Recipe;
+
+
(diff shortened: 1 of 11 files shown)


cm0x4D:
@Baughn: In order to get it to compile on travis without errors, you might need to select JDK8 like this:

```
jdk:
  - oraclejdk8
```

in .travis.yml.

Seems Java 7 is still the default on travis...

jrddunbr:
Related: https://github.com/age-series/ElectricalAge/pull/275

I found that there might be a bug in the removal code here, when I converted it to Kotlin. It only tries to check the Macerator recipes.

src/main/java/mods/eln/integration/minetweaker/utils/RemoveRecipe.java
```diff
@@ -0,0 +1,36 @@
...
+	
+	public RemoveRecipe(MinetweakerMachine machine, IIngredient input) {
+		super(machine);
+		this.input = input;
+	}
+	
+	@Override
+	public void apply() {
+		toRemove = new ArrayList<Recipe>();
+		for (Recipe r : Eln.instance.maceratorRecipes.getRecipes()){
```
jrddunbr, line 23: This is probably a bug

PR #480, opened 2016-07-07
Add WAILA

bloxgate:
Adds WAILA support for all transparent nodes. Includes a configurable "Easy Mode" that shows more info on a few devices.

Also adds a plotBuckets and a readPrivateDouble function to Utils.

---

This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/480)

### DIFF
--- a/src/main/java/mods/eln/Eln.java
+++ b/src/main/java/mods/eln/Eln.java
@@ -10,8 +10,7 @@ import cpw.mods.fml.common.network.simpleimpl.SimpleNetworkWrapper;
 import cpw.mods.fml.common.registry.EntityRegistry;
 import cpw.mods.fml.common.registry.GameRegistry;
 import cpw.mods.fml.relauncher.Side;
-import mods.eln.achievepackets.AchievePacket;
-import mods.eln.achievepackets.AchievePacketHandler;
+import mods.eln.packets.*;
 import mods.eln.cable.CableRenderDescriptor;
 import mods.eln.client.ClientKeyHandler;
 import mods.eln.client.SoundLoader;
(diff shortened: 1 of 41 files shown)


cm0x4D:
@bloxgate Is this ready to be merged? So it could be part of the next release...

bloxgate:
@cm0x4D Yes. Last I checked it was working on all the machines, though I'd feel better if someone took a look and made sure I hadn't missed anything.

src/main/java/mods/eln/Eln.java
```diff
@@ -297,11 +296,16 @@
...
 	double stdHalfLife = 2 * Utils.minecraftDay;
 
+	public static boolean wailaEasyMode = false;
+
 	@EventHandler
 	public void preInit(FMLPreInitializationEvent event) {
 		
-		achNetwork = NetworkRegistry.INSTANCE.newSimpleChannel("achChannel");
-		achNetwork.registerMessage(AchievePacketHandler.class, AchievePacket.class, 0, Side.SERVER);
+		elnNetwork = NetworkRegistry.INSTANCE.newSimpleChannel("eln");
```
cm0x4D: @bloxgate Renaming this channel has no side effects?

cm0x4D (commented):
LGTM, I am just not sure that renaming the network channel does not have any unwanted side effects...

Baughn:
Renaming the channel has the side-effect that you won't be able to connect to a server running this patch with a client that doesn't, or vice versa. That's not a big deal.

I'm concerned about possible name clashes, with it being just three letters, but I suppose it's still unlikely.

src/main/java/mods/eln/Eln.java
cm0x4D: Maybe using "electricalage" or "electrical-age" in place of "eln"?

bloxgate: As mentioned below, as long as someone doesn't try using a different version than the server has it should be fine.

bloxgate: @cm0x4D I've changed the channel name to electrical-age

Baughn (approved):

cm0x4D:
Reviewed 41 of 43 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/electrical-age/electricalage/480)_

cm0x4D:
Merged, thanks @bloxgate for your contribution!

PR #494, opened 2016-07-14
Logic gates recipes & in-game infos

cm0x4D:
Ready to merge.


cm0x4D:
@Electrical-Age/devs As I am rather the creative player, the recipes might not be top notch, but they will probably do the trick, if someone has improvements feel free to change them...

PR #502, opened 2016-07-26
Old-school lead fuses

cm0x4D:
Not ready to merge, I need to add:
- [x] Texture for model
- [x] Fuse items (One for each cable type)
  - [x] Item icons
  - [x] Item code
- [x] Fuse code
  - [x] Descriptor
  - [x] Element
  - [x] Render
  - [x] Process
- [x] Recipes
  - [x] Fuse holding
  - [x] Fuse items

---

This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/502)


Baughn:
So these would be single-use fuses?

They should be cheaper than the breakers, then. Let me think... how's this?

```
WWW
C C
WWW
```

Where W is wood, for the 50V (cheapest) version, C is copper. The 200V version might use plastic instead. I don't think I've ever seen a non-resettable 800V or higher fuse, and I suspect it wouldn't be used; these are for early-game, obviously.

A single piece of uninsulated lead wire goes in the UI. I don't think we have those yet, but that should be easy.

To avoid usefulness inversion, the electrical breakers ought to acquire a signal input.

cm0x4D:
> So these would be single-use fuses?

Yes and no. It is similar to lamp socket and lamp items. There is a generic case for the fuses and there are items (the lead fuses) you can put into the case.

They should be simpler to use than the breaker and a lot cheaper to craft. With the disadvantage that you have to change them if they trigger.

> I don't think I've ever seen a non-resettable 800V or higher fuse, and I suspect it wouldn't be used; these are for early-game, obviously.

My idea is that they trigger much slower than the breaker, so the cables get pretty hot before the fuse actually triggers. So they are much more "user friendly" as the breaker. I would like to provide fuse items for maximal currents of 50V, 200V and 800V cables and maybe some more in between...

AMIDIBOSS:
> Texture for model
>  Fuse items (One for each cable type)
>  Item icons

I can help with these

AMIDIBOSS:
Fuses are more rated on amps, than on volts

AMIDIBOSS:
Will a multimeter be able to check fuse?

cm0x4D:
Yes and no. Multimeter shows the actual amperage and temperature, but the fuse itself has an indicator if it is alright or not.

AMIDIBOSS:
What'll be the trigger temp. of the fuse?

cm0x4D:
Well, yes it is the temperature of the fuse that will actually burn out the fuse.

AMIDIBOSS:
The fuse will have only voltage rating? 
Or will there be something like current or power rating?  Is there any other way to limit current? 
Can you cosider adding fuse slot for 200v machines? Or switches with fuse slot?

cm0x4D:
> The fuse will have only voltage rating? 
> Or will there be something like current or power rating?

They are rated on current, but there exists at the moment a fuse for each cable type. Basically the fuse wire is dimensioned to burn at the right current, but it is actually a thermal process that burns the fuse.

> Is there any other way to limit current?

There is the breaker which can be used to protect cables and equipment. Once the current is surpassed or the voltage is outside the parameters, it immediately opens it's contact.

If you want to limit the current without cutting you could probably build a current limiter using a signal processor and a rheostat or you do some PWM (pulse width modulation) with a relay in order to limit the current. But remember, limiting the current causes always a voltage drop at the consumers...

> Can you cosider adding fuse slot for 200v machines? 
> Or switches with fuse slot?

Probably not, never say never, but there are not that much devs left and we have to think twice what we add and what not...

Baughn:
You can also limit current with an inductor with a relay in front of it. Although that only works because relays don't have a voltage watchdog, I'm okay with that, on the basis that the proper buck circuit just doesn't work.

May have to increase the max inductor size, though. 100H just isn't enough for a voltage difference of 3200V.

cm0x4D:
Reviewed 1 of 9 files at r3, 14 of 15 files at r4, 7 of 10 files at r5, 3 of 3 files at r6.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/electrical-age/electricalage/502)_

PR #516, opened 2016-09-24
Tachometer: Use rad/s, and sane defaults.

Baughn:
So far every mechanical device runs optimally at 800 rad/s, and explodes at 1000.

---

This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/516)


Baughn:
@Dolu1990: You can hit 'merge', then do the release. :)

PR #519, opened 2016-10-02
Feature/powerline

Baughn:
All right, enough already. This code has been stable for over a year; even if I'm still working on it, it's time to let people use it. I've waited long enough for someone to fix the Kotlin converter, and doing the conversion by hand is error-prone.

What we've got below is a piece of code that's the first serious modification I made to EA, but has also been _thoroughly_ battle-tested. It's solid, and the rewrite I apparently need to do to get it up to my current expectations will take a while, as will testing post-rewrite, so there's really no reason to wait. I'm certain I can maintain on-disk compatability.

One thing that's currently missing, though. The transformer poles should have a "foot"; a non-rotating square bit at the bottom, which cables can be rendered as going into. So:
- [x] Implement that foot. Draw cables.
- [x] Stop registering the "downlink" and "grid transformer", which are still WIP. They're not craftable, but they take space in NEI. Don't get rid of the code, though.
- [ ] Fix the GL_CULL_FACE problem.
- [x] Make cables draw if placed vertically.

If anyone wants to do either of those, let me know. Otherwise I'll try to get to it this upcoming week.

---

This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/519)


metc:
@Baughn you should update the [.travis.yml](https://github.com/Electrical-Age/ElectricalAge/blob/develop/.travis.yml) configuration file to use javac 8.

It is also possible to setup a build matrix if we want to target several Java versions.

Baughn:
@metc - This code depends on Java 8 to work, although not for any very strong reasons. On the other hand, setting that JVM limitation also improves Kotlin performance a bit. And a lot of mods (such as AE2) already depend on 8 anyhow.

metc:
I do not use Mineracft mods that requires Java 8 to work. I have nothing against switching to java 8.
From a developer point of view, Java 8 is less plain full, but I do not know if every Minecraft player has a 1.8 JVM installed and if this will be a limitation or not...

AMIDIBOSS:
So, where we are going to get a build that has this famous "code" implemented?
Malybe you'll aolso update the tutorial world? (Unless it's abondoned)

Baughn:
PTAL?

The only remaining issue I can see is the GL_CULL_FACE inefficiency, and it isn't something I've actually noticed mattering, although we've used quite a lot of power poles. Since I'll probably need to redo this code entirely for 1.10 anyhow, I'd rather not worry about it until then.

...and the "art". But I don't think I can fix that easily.

cm0x4D:
LGTM, thanks @Baughn, very nice. Maybe the models could be a little more textured, for example the transformer on the pole. I would do that if you like.

Baughn:
Please do. I'm _very_ slowly improving, but I'm still no good at this.

cm0x4D:
Reviewed 19 of 28 files at r1, 5 of 9 files at r2, 5 of 5 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/mods/eln/gridnode/GridRender.java, line 24 at r3](https://reviewable.io:443/reviews/electrical-age/electricalage/519#-KXbUh_chRDEoMtUPx0Y:-KXbUh_dyaSRfdYkdXnd:b-bz63xa) ([raw file](https://github.com/electrical-age/electricalage/blob/ea1f612ba10a9d154dc03eaf9facd636f1142664/src/main/java/mods/eln/gridnode/GridRender.java#L24)):*
> ```Java
>     private final GridDescriptor descriptor;
>     private final ResourceLocation cableTexture;
>     private ArrayList<Catenary> catenaries = new ArrayList<Catenary>();
> ```

Could be final?

---

*[src/main/java/mods/eln/gridnode/downlink/DownlinkElement.java, line 22 at r3](https://reviewable.io:443/reviews/electrical-age/electricalage/519#-KXbXKEnQAaYHokXjUsJ:-KXbXKEnQAaYHokXjUsK:b-oyuf3j) ([raw file](https://github.com/electrical-age/electricalage/blob/ea1f612ba10a9d154dc03eaf9facd636f1142664/src/main/java/mods/eln/gridnode/downlink/DownlinkElement.java#L22)):*
> ```Java
>  * Created by svein on 25/08/15.
>  */
> public class DownlinkElement extends GridElement {
> ```

A lot of these fields could be final, method parameters too. But that's not really a problem at all I guess.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/electrical-age/electricalage/519)*

PR #536, opened 2016-11-20
WAILA for GhostBlocks

cm0x4D:
see #533 

not yet ready to merge, I still have to fix the issue with the icon and name not shown.

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/536)


cm0x4D:
@bloxgate I tried to get myself into the WAILA stuff. I managed to get basic WAILA to work for ghost blocks, but the icon and the name are not working yet :(. 

Do you mind to have a look at the code and see if this might work and do you have an idea what I would have to do in order to get the icon and name working?

The wind turbine or water wheel works quite good to test the WAILA integration with ghost blocks

bloxgate:
If I remember correctly the icon can be overridden using the getWailaStack method. I don't know how to do the name, but the rest of your code looks like it should work fine.

Edit: You may have to register a stack provider for the getWailaStack method to work, but I'm not entirely sure.

cm0x4D:
Thanks you @bloxgate, but seems not to work, the only method called by WAILA is getWailaBody(), all other methods get never ever called, so I have no chance to change the output...

cm0x4D:
Ahh, got it. I registered it as body provider, but I should have taken something else...

cm0x4D:
Ok then. Should be ready to merge, if someone could have a look, I did not test it extensively...

src/main/java/mods/eln/integration/waila/GhostNodeWailaProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+    private fun getGhostData(coord: Coordonate): GhostNodeWailaData? {
+        var ghostData: GhostNodeWailaData? = null
+
+        if (Minecraft.getSystemTime() - GhostNodeWailaProvider.updateTime > 2000) {
+            Eln.elnNetwork.sendToServer(GhostNodeWailaRequestPacket(coord))
+            GhostNodeWailaProvider.updateTime = Minecraft.getSystemTime()
+        }
+        try {
+            ghostData = WailaCache.ghostNodes.get(coord)
+        } catch(e: Exception) {}
```
Baughn: This is a really wide catch. Could you possibly catch something more specific?

Switching to Guava may obsolete this comment.

src/main/java/mods/eln/integration/waila/GhostNodeWailaProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+@Optional.Interface(iface = "mcp.mobius.waila.api.IWailaDataProvider", modid = "Waila")
+class GhostNodeWailaProvider(private val transparentNodeHandler: TransparentNodeHandler): IWailaDataProvider {
+    companion object {
+        var updateTime = Minecraft.getSystemTime()
+    }
+
+    private fun getGhostData(coord: Coordonate): GhostNodeWailaData? {
+        var ghostData: GhostNodeWailaData? = null
+
+        if (Minecraft.getSystemTime() - GhostNodeWailaProvider.updateTime > 2000) {
```
Baughn, line 48: Rather than building your own cache implementation, prefer to use Guava's. API docs here: http://google.github.io/guava/releases/20.0/api/docs/, or more concretely here: http://google.github.io/guava/releases/20.0/api/docs/com/google/common/cache/Cache.html

The `get` method would automatically load the key if nonexistent, which is bad (it'd block), but you can combine `getIfPresent` and `refresh` (from LoadingCache) to get the desired behaviour.

You're already using it elsewhere.

src/main/java/mods/eln/integration/waila/GhostNodeWailaProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+        try {
+            ghostData = WailaCache.ghostNodes.get(coord)
+        } catch(e: Exception) {}
+
+        return ghostData
+    }
+
+    override fun getWailaBody(itemStack: ItemStack?, currenttip: MutableList<String>?, accessor: IWailaDataAccessor?,
+                              config: IWailaConfigHandler?): MutableList<String>? {
+        val pos = accessor!!.position
```
Baughn: If you're sure that accessor is non-null, then remove the ? from the parameter. Kotlin allows you to do that, for Java functions without nullability annotations. (And if it errors out then, then clearly you need to check.)

src/main/java/mods/eln/integration/waila/GhostNodeWailaProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+        val ghostData = getGhostData(Coordonate(pos.blockX, pos.blockY, pos.blockZ, accessor.world))
+
+        if (ghostData != null) {
```
Baughn: This can more idiomatically be written such:

```kotlin
return if (ghostData != null) {
  <same code here, but remove the 'return'>
} else {
  currenttip
}
```

src/main/java/mods/eln/integration/waila/GhostNodeWailaProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+            pos.blockZ = realCoord.z
+
+            return transparentNodeHandler.getWailaBody(itemStack, currenttip, accessor, config)
+        } else {
+            return currenttip
+        }
+    }
+
+    override fun getWailaStack(accessor: IWailaDataAccessor?, config: IWailaConfigHandler?): ItemStack? {
+        val pos = accessor!!.position
```
Baughn: Ditto. I won't generally say "Do the same thing here", though.

src/main/java/mods/eln/integration/waila/GhostNodeWailaProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+
+        if (ghostData != null) {
+            return ghostData.itemStack
+        } else {
+            return null
+        }
+    }
+
+    override fun getWailaTail(itemStack: ItemStack?, currenttip: MutableList<String>?, accessor: IWailaDataAccessor?,
+                              config: IWailaConfigHandler?): MutableList<String>? {
```
Baughn: Whole thing can be idiomatically rewritten to
```kotlin
override fun getWailaTail(itemStack: ItemStack?, currenttip: MutableList<String>?, accessor: IWailaDataAccessor?, config: IWailaConfigHandler?) = currenttip
```

In general, prefer to write single-expression functions that way.

src/main/java/mods/eln/integration/waila/GhostNodeWailaProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+        return currenttip
+    }
+
+    override fun getNBTData(player: EntityPlayerMP?, te: TileEntity?, tag: NBTTagCompound?,
+                            world: World?, x: Int, y: Int, z: Int): NBTTagCompound? = null
+
+    override fun getWailaHead(itemStack: ItemStack?, currenttip: MutableList<String>?, accessor: IWailaDataAccessor?,
+                              config: IWailaConfigHandler?): MutableList<String>? {
+        val itemStack = getWailaStack(accessor, config)
+        if (itemStack != null) {
```
Baughn: Whole function could be rewritten as a single expression.

Baughn (changes requested):

src/main/java/mods/eln/integration/waila/GhostNodeWailaProvider.kt
cm0x4D: The method get calls the method load of the cache, which does in our case only send the request packet to the server. So it is not really blocking, but still kind of a hack. That is why get throws an exception - the load method does not return the desired value.

src/main/java/mods/eln/integration/waila/GhostNodeWailaProvider.kt
cm0x4D: We are actually using Guava cache:
ghostNodes is actually LoadingCache<Coordonate, GhostNodeWailaData>

But using Guava's cache the way it was meant will make it obsolete...

src/main/java/mods/eln/integration/waila/GhostNodeWailaProvider.kt
cm0x4D: Thanks, I mostly forget that. Does not matter if the exception comes from the implicit or explicit unwrapping...

src/main/java/mods/eln/integration/waila/GhostNodeWailaProvider.kt
cm0x4D: The cache is not really adapted to our needs, so I guess the actual solution is the simplest. I do not want to save futures in a list in order to wait to call them when the response from the server has arrived nor do I want to spawn a thread for each pending message. Its a hack to use something a little different than it's intention, but it works.

Baughn (approved):

AMIDIBOSS (approved):

PR #543, opened 2016-12-02
Waila for ghost blocks of sixnode blocks

cm0x4D:
Not yet ready to merge, 

- [ ] Display the content for the represented sixnode block for ghost block.

@bloxgate: I would need to get the waila data for the represented sixnode [here](https://github.com/Electrical-Age/ElectricalAge/blob/bugfix/streetlight-waila/src/main/java/mods/eln/integration/waila/GhostNodeWailaProvider.kt#L67), but I do not know how to do it, any idea?

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/543)

### DIFF
--- a/src/main/java/mods/eln/integration/waila/GhostNodeWailaData.kt
+++ b/src/main/java/mods/eln/integration/waila/GhostNodeWailaData.kt
@@ -1,6 +1,9 @@
 package mods.eln.integration.waila
 
 import mods.eln.misc.Coordonate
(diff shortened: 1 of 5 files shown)


cm0x4D:
@Baughn If you like to release a new version please merge this first. 
Even if it is not finished yet, it avoids that the server crashes when looking on sixnodes with ghostblocks.

cm0x4D:
I merge now to fix the bug, I will improve the sixnode ghost block information in the sixnode branch...

PR #545, opened 2016-12-04
WAILA integration for sixnode blocks

cm0x4D:
- [x] I have still to fix the problem with the sixnode ghost blocks before this can be actually merged.
- [x] We should add custom waila content for different sixnode blocks.

@bloxgate If you like to help adding these it would be very welcome...

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/545)


cm0x4D:
@bloxgate I modified the cache behavior for sixnode and ghost blocks and changed the size to about 20 which imho should be enough, as the info is reloaded anyway each 2 seconds. Can I adapt your code for the transparent node info too?

bloxgate:
Go ahead.

cm0x4D:
@bloxgate There is support to print values already integrated into the mod. You can use Utils.plotXY() to print well formatted values. This has the advantage that it prefixes the unit with metric prefixes like (k, m, u, ...)

bloxgate:
Some of the machines use it. Only found out about that feature a bit of the way into writing all the original getWaila()'s

cm0x4D:
@Electrical-Age/devs If someone mind to have a look, should be ready to merge.

PR #549, opened 2016-12-09
DCDC Converter replacement for Transformator

cm0x4D:
TODO:
- [x] Finish model and textures
- [x] Rename Transformator to DCDC converter
- [x] Replace transformer icon with DC/DC icon

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/549)


AMIDIBOSS:
And what about utility pole w transformer? Will they get new textures too? And grid transformer! Will it get new industrial look instead of ugly model that it has?

cm0x4D:
@Baughn Replacing the transformer with a DCDC converter conflicts a little bit with the recipe for the  pole with transformer, as the recipe actually needs a transformer. For the moment I replaced it with the DC-DC converter, but imho this is not a very good solution. Either we add a transformer item just for the poles or we decide that the poles use DCDC converters too. On the one side it feels a little strange to have a DCDC converter on a pole, but on the other side if we would actually rebuild the real power grid from ground up today , it would probably be DC...

What do you think?

AMIDIBOSS (approved):

AMIDIBOSS:
I think it's just minecraft. If the floating islands aren't strange, why would polemount DC converters do? And if you have barrel-cased transformer, why can't you have barrel-cased  converter? No need to change model!  Just tweak name and crafting a bit, and everything's fine! About real power grid, it might be DC, because of solar panels. On the other side, DC is hard to transfer and distribute. When it's going to get merged?

AMIDIBOSS (approved):

Baughn:
@cm0x4D It's a DC grid, so a DCDC converter is what the poles would be using. It may seem a little odd, but... I have no problem with that idea.

Baughn (approved):

cm0x4D:
@Baughn Ok. I replaced the transformer on the pole with an industrial box pretending to contain a DC DC converter. Please have a look to the model in game and give me your ok for the changes on the model.

Baughn:
http://i.imgur.com/Sssowta.png

@cm0x4D 👍

AMIDIBOSS (approved):
I like the model

cm0x4D:
ready to merge

Baughn:
I've tested this on E14, and it works fine. Merging.

PR #550, opened 2016-12-12
fuel heat furnace

cm0x4D:
The basic idea is that this fuel burner takes fuel of any kind (oil, gaz, ...) and produces heat that can be used for the heat turbines. Should be compatible with forge liquids and should offer the possibility to pipe the fuel into.

I imagine that there is the basic block that can be equipped with burners with different heat power outputs.

Missing:
- [x] Textures
- [x] Burner items
- [x] Code
- [x] Sample
- [x] Recipe for basic block and burner items
- [x] Ensure that the inventory is synched to the client when the element is created

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/550)


cm0x4D:
Inspiration for model:
heizung

AMIDIBOSS (approved):

bookerthegeek (approved):
Looks nice. ☺

AMIDIBOSS (approved):

AMIDIBOSS (approved):

AMIDIBOSS (approved):
There will be three sisez?

cm0x4D:
> There will be three sisez?

Yes and no. There are 3 different burner items that can be installed. It is similar to the burning chamber of the stone heat furnace.

cm0x4D:
@Baughn I would need now to get the burn energy per Bucket or mB of a fuel. I guess that this is a common task, so it would actually make sense to put this into the FuelRegistry. For example the heat furnace would convert 100% of that energy into heat, whereas other machines like the fuel generator would be less efficient, but still we would need to determine the amount of energy that can be extracted from the fuel. I mean oil should not have the same energy density as gasoline.

I do not know that much other mods, but do you have an idea how a Bucket or mB relates to real world units like liters? Then I could use the real word energy density of those fuels. Or do you have another idea how to do this correctly? Maybe there is an interface I do not know about...

Last thing: I made some changes to the fluid code, could you please have a look end let me know if this is alright for you?

Thanks

AMIDIBOSS (approved):

AMIDIBOSS (approved):

AMIDIBOSS (approved):
Why is there "Fire Alarm sound removed" noted, it haven't been touched?

cm0x4D:
@Baughn Can you check the recipes and give some feedback? I am really not an expert on how expensive these items should be.

src/main/java/mods/eln/Eln.java
```diff
@@ -787,11 +797,14 @@ public void load(FMLInitializationEvent event) {
 		recipeCompressor();
 		recipePlateMachine();
 		recipemagnetiser();
+		recipeFuelBurnerItem();
 
 		recipeECoal();
 
 		recipeGridDevices(oreNames);
 
+
```
Baughn: Try not to add excessive whitespace.

src/main/java/mods/eln/Eln.java
```diff
@@ -2682,6 +2695,15 @@ void registerHeatFurnace(int id) {
 			transparentNodeItem.addDescriptor(subId + (id << 6), desc);
 		}
 
+		{
+			subId = 1;
+			name = TR_NAME(Type.NONE, "Fuel Heat Furnace");
+
+			FuelHeatFurnaceDescriptor desc = new FuelHeatFurnaceDescriptor(name,
+				obj.getObj("FuelHeater"), new ThermalLoadInitializerByPowerDrop(780, -100, 10, 2));
```
Baughn, line 2701: Do you know how the ThermalLoadInitializerByPowerDrop works? If so, I'd adore if you could add a bit of documentation.

src/main/java/mods/eln/Eln.java
```diff
@@ -4799,6 +4821,15 @@ void registerPortableItem(int id) {
 		}
 	}
 
+	private void registerFuelBurnerItem(int id) {
+		sharedItemStackOne.addElement(0 + (id << 6),
+			new FuelBurnerDescriptor(I18N.TR_NAME(Type.NONE, "Small Fuel Burner"), 5000 * fuelHeatFurnacePowerFactor, 2, 1.6f));
```
Baughn, line 4824: "5000" is a magic number. How did you come up with it? Can you put that logic in the code?

src/main/java/mods/eln/Eln.java
```diff
@@ -5515,6 +5546,13 @@ void recipeHeatFurnace() {
 				Character.valueOf('i'), findItemStack("Copper Thermal Cable"),
 				Character.valueOf('I'), findItemStack("Combustion Chamber"));
 
+		addRecipe(findItemStack("Fuel Heat Furnace"),
```
Baughn, line 5547: With liquid fuel, the main cost is the fuel itself and I'd probably default to a realistic-ish but reasonably simple recipe. That said...

Some sort of regulation for amount-of-fuel-accepted seems like it'd make sense, but these don't need power, so a motor wouldn't fit. If they're sufficiently powerful (How many 200V turbines can one of these power?), then having it use a bit of electricity could make sense -- though we're generally treating signals-level power as free, and given its nature there could be a small peltier in there anyway.

Er, balancing, right. This depends on how powerful they really are, but assuming this can replace at least two heat furnaces, I'd suggest adding a cauldron in the middle and swapping out one of the side iron ingots for a motor. You may or may not add code to make it actually need power; that's not such a big deal.

If the smallest fuel burner *doesn't* have the capacity to replace at least one fully-upgraded heat furnace -- then it should. It's generally easier to get charcoal than fuel, and it's definitely easier to handle, so players should get some benefit from going the liquid fuel path.

src/main/java/mods/eln/Eln.java
cm0x4D: No idea, I just copied it from the Heat Furnace. It kind of calculates the thermal parameters to arrive to the given timings, that is all I know.

src/main/java/mods/eln/Eln.java
cm0x4D: Well it's 5 kW of heat power. Thats a design decision.

src/main/java/mods/eln/Eln.java
cm0x4D: > I'd suggest adding a cauldron in the middle and swapping out one of the side iron ingots for a motor.

Good point. I will do that.

> You may or may not add code to make it actually need power; that's not such a big deal.

For simplicity I would prefer to not to, there are already the fuel, the thermic and the control connections...

> f the smallest fuel burner doesn't have the capacity to replace at least one fully-upgraded heat furnace -- then it should.

Ok, I will check that.

Thanks for the input.

src/main/java/mods/eln/Eln.java
Baughn: In that case, perhaps rename powerFactor to WattsPerBucket (or milliBucket).

cm0x4D: Oh, this factor can be used to balance the fuel heat furnace power in the ELN config file like for a lot of other machines, it has nothing to do with the fuel consumption.

src/main/java/mods/eln/Eln.java
cm0x4D: The smallest fuel burner has a power of 5 kW, the heat furnace has 1-2kW of heat power (depends the number of combustion chambers), so the smallest burner has more than twice the power of the simple heat furnace. Imo this is ok.

Baughn:
LGTM.

I'm rebasing the 'turbinesound' branch on top of this, and will be using your sound-looping code. Let's see how it works~~

cm0x4D:
Good idea, some testing might be quite useful. Maybe on a multliplayer server...

---

Review status: 0 of 23 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/main/java/mods/eln/Eln.java, line 806 at r5](https://reviewable.io:443/reviews/electrical-age/electricalage/550#-KYz2-OPC7UnpfDq3APc-r5-806:-KaI11tPyF0kHw9K9NFq:b-ybavzs) ([raw file](https://github.com/electrical-age/electricalage/blob/86858c7f207d719f180ca4c0b02a2bd7f1a8131a/src/main/java/mods/eln/Eln.java#L806)):*
<details><summary><i>Previously, Baughn (Svein Ove Aas) wrote…</i></summary><blockquote>

Try not to add excessive whitespace.
</blockquote></details>

Ok. We should once we do not have that much PR clean up the code a little. IDEA can do this for us ;-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/electrical-age/electricalage/550#-:-KaI0qbM_nlWmZtHKQIS:bdw41a)*

Baughn:
Now, where would I find one of those. :P

src/main/java/mods/eln/Eln.java
```diff
@@ -6640,6 +6678,29 @@ void recipemagnetiser() {
...
+			" Cc",
+			"   ",
+			Character.valueOf('C'), findItemStack("Combustion Chamber"),
+			Character.valueOf('c'), findItemStack("Copper Thermal Cable"));
+
+		addRecipe(findItemStack("Medium Fuel Burner"),
+			"   ",
+			" Cc",
+			"   ",
+			Character.valueOf('C'), findItemStack("Combustion Chamber", 2),
```
Baughn: This doesn't work; you can't require multiple chambers in the same slot. Try spreading them out.

Baughn (changes requested):

src/main/java/mods/eln/Eln.java
cm0x4D: Ok, it is fixed now.

src/main/java/mods/eln/Eln.java
```diff
@@ -6689,15 +6689,15 @@ void recipeFuelBurnerItem() {
 		addRecipe(findItemStack("Medium Fuel Burner"),
 			"   ",
 			" Cc",
-			"   ",
-			Character.valueOf('C'), findItemStack("Combustion Chamber", 2),
+			" C ",
```
Baughn, line 6692: Just a nit, but it'd be nicer to put all these on a line.

Baughn:
I've tested on Erisia, and it works just fine. Merge when desired.

Baughn (approved):

src/main/java/mods/eln/Eln.java
cm0x4D: The 3D model of the burner is rather high than long, so I thought this recipe matches better the actual appearance of the model... But I can change it before merging. What do you think?

cm0x4D:
> I've tested on Erisia, and it works just fine. Merge when desired.

That was the intention ^^. No problems with the looped sounds? It would be awesome if this works reliable. Next step would be to port all looping sounds to the new API...

Baughn:
There are some very slight glitches, which will take careful doctoring of the sound files to eliminate, but the code works perfectly. It's about 1,000% better than it was.

src/main/java/mods/eln/Eln.java
Baughn: Point. I didn't consider its appearance.

PR #571, opened 2016-12-27
Fix recipes when aluminium isn't available

nikola3244:
Problem:
Generator and Steam Turbine can't be crafted if aluminium ingot and block do not exist.

Solution:
Check if `ingotAluminium` and `blockAluminium` exist; if they do, use them; else use `ingotIron` and `blockIron`.

P.S.
It's untested, but based on some common sense (and example from #L5579), it should work.

### DIFF
--- a/src/main/java/mods/eln/Eln.java
+++ b/src/main/java/mods/eln/Eln.java
@@ -5538,23 +5538,43 @@ public class Eln {
 				Character.valueOf('E'), findItemStack("Medium Voltage Cable"),
 				Character.valueOf('H'), findItemStack("Copper Thermal Cable"),
 				Character.valueOf('m'), findItemStack("Advanced Electrical Motor"));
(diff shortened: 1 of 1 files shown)


cm0x4D:
Hello @nikola3244, thanks for the proposed fix. I added a method to handle such situations in PR #573 because imho it is not a good idea to duplicate the recipe code which could lead to errors once we copy or modify recipes.

nikola3244:
Hey @cm0x4D, your code is much better!

To be honest, I've thought that my code isn't great, but since I don't have much experience with Java I just followed an existing example that was below the proposed fix. 

Anyway, good points and a great solution.

cm0x4D:
Most people are just yelling that something does not work, you provided directly a solution to the problem you discovered. Thank you and keep this spirit!

PR #578, opened 2016-12-28
Add various turbine sounds

Baughn:
- [X] Generator whine
- [x] Gas turbine running
- [x] Steam turbine running
- [x] Fix the static
- [x] Find replacement sounds

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/578)


Baughn:
A problem: At present, there's a kind of static to the sound.

I suspect this is due to the arbitrary pitch shifts.

ReikaKalseki:
Permission **not** given. If you want to use them as temporary never-released "as a test to make sure the playback system works", sure, but identical files in two different mods is going to cause headaches for both of us.

AMIDIBOSS:
There will be pepole saying "Electrical age copied rotary craft", others will say "Rotary craft copied Electrical age", Both of you will get spammed: "You coped from RC/ELN". Pure nightmare.

Baughn:
No? We discussed this once before, and it doesn't actually sound very similar...

Ah well. I'll find some other samples.. somewhere. Besides, the playback system doesn't actually work. I can't seem to get rid of that crackle.

AMIDIBOSS:
If you need i can record my basement generator, sounds very like a gas turbine :)

Baughn:
@AMIDIBOSS That would be appreciated, but at present the chance of getting to use those samples seems to be shrinking.

I can *not* figure out what the cause of this crackle is. Eln doesn't actually do anything terribly unusual, it's just calling playSound... but crackle there is nonetheless.

cm0x4D:
@Baughn The sound system is a nightmare. You need your samples to have a fade-in and fade-out (equal power ramps) in order to be able to loop them without to many problems. I did already quite a bunch of these samples (almost all), so if I have the raw sample, I can adapt it to sound acceptable when looped.

AMIDIBOSS:
Why don't we use gitter for this?

Baughn:
@AMIDIBOSS Probably because I don't usually have gitter open. I'm always on IRC, though.

ReikaKalseki:
Seamless looping does not necessarily require a fade on either end; what it does require is that the waveform for the first few tenths of a second line up with that of the last few. I usually modify files in Audacity - with the pencil tool, editing individual samples - for this.

Baughn:
Also see ESRever's post on http://forums.rpgmakerweb.com/index.php?/topic/40517-using-audioblocks-for-music-resources/ -- that's a really nice list of resources.

And sorry, Reika, I don't think I have the patience to manually edit 4800 samples to ensure overlap. This should be automatable... actually, I think I know how to do so.

ReikaKalseki:
4800? More like 20.

cm0x4D:
You can try these. 
[Turbines & Generator Samples.zip](https://github.com/Electrical-Age/ElectricalAge/files/679224/Turbines.Generator.Samples.zip)

They are made quite similar to the existing ones with ramps on both ends to simplify looping. Afaik we do not have sample accurate looping support (yet), so even if the samples would be perfect, we would still have some static due to the fact that the samples do not glue perfectly.

The Gas turbine sample comes from Domi1993 via mail, the Generator comes from [here](https://www.freesound.org/people/jameswrowles/sounds/327565/) and the steam turbine sample from [here](https://www.freesound.org/people/Sea%20Fury/sounds/54638/). The last would need an actual attribution (Is this a problem?).

The Gasturbine sample needs to be looped each 1.25s and the others each 1.75s.

Baughn:
I can't get Minecraft to cooperate, though. The sound system keeps failing, cutting off *all* sound from the game...

That's probably some limitation I'm not aware of, but what can you do? Probably the best solution is to write my own. That'll take a while, and I don't think I should work on that before at least trying a bit harder on the 1.10 port. ...if I can get *that* working...

I'll give it a few more prods. See if there's any simple fixes.

Domi1993:
What about the startup sound, i think we dont need a specific sound for starting because.

The turbine startup is dependent of the setup the players are making. If a flywheel is used ....

The motor is making a loud noise for spin up the gasturbine, this in combination with the gasturbine sound sounds like a startup.

I'll build a test powerplant on my ELN World, and upload a Video on youtube. Next week

Baughn:
@Dolu1990 After some testing on Erisia, I tend to agree; no specific startup sound is needed.

Also, although I'll need to make SoundCap an optional dependency — and add a config option to turn it off — I think this code should be stable enough to be usable without massive changes. Things break mostly if there's TPS jitter, or network... which isn't ideal, mind you.

The remaining issues are:
- Replace the sounds with something Reika doesn't mind if I use. Ehh... I don't suppose he'd be fine with just a pitch shift? No? Meh, I'll figure something out.
- Move the looping to client-side. Which I'll do *later*, it's not a blocker.

@ReikaKalseki, you have the same problem with looping sounds. Would you be interested in joining forces on this one? Well, basically I'm wondering if I ought to make client-side sound looping into a separate mod.

Domi1993:
@Baughn

But the sounds are now integrated so i can test it?

ReikaKalseki:
When I want to loop a sound, I just play the sound every N ticks, where N is the length of the sound in 20ths of a second. As for the actual sound file, I just ensure it is approximately an integer number of 20ths of a second and that it has matching sample waveforms for the 15-20 or so samples on either end.

Baughn:
That's what everyone does, and it fails quite a lot.

If the TPS of the server falls below 20, then the sound will fade in and out. If there are any network problems at all, then the client may receive the "Play this sound" requests bunched up. The only way to do this *reliably* is to push as much of the logic as possible client-side.

Which means doing the same thing you're suggesting, just... on the client, in a separate thread, and on a timer rather than based on ticks.

cm0x4D:
Looping samples on the client (TransparentNodeRender) is not to complicated when using either the ISound or the ITickableSound interfaces and the Minecraft sound handler: https://github.com/Electrical-Age/ElectricalAge/pull/550/commits/92279038597402c2a160e2c91eae158e037c3c1f

The samples even do not need to match any tick interval or whatsoever. The code is quick and dirty and I had not that much time to test it, but basically it already worked for one single fuel heat furnace I placed into the world...

Baughn:
That's a lot easier than expected. Do you think you could to it for the SoundLooper class?

cm0x4D:
>  Do you think you could to it for the SoundLooper class?

No. I agree with you that all this sample looping should happen exclusively on the client side. I created a new class [LoopedSound](https://github.com/Electrical-Age/ElectricalAge/pull/550/commits/6d0c4e80221f7c8ec8d197aadd537bb4ee51e715#diff-8e2481edabd839f9997f4c61ef2daabf) and added the API to add and remove looped sounds to the [TransparentNodeElementRender](https://github.com/Electrical-Age/ElectricalAge/pull/550/commits/6d0c4e80221f7c8ec8d197aadd537bb4ee51e715#diff-0916751fbe891927f08e834248d8cff1) base class. It is imo very simple to use. An example is [here](https://github.com/Electrical-Age/ElectricalAge/pull/550/commits/6d0c4e80221f7c8ec8d197aadd537bb4ee51e715#diff-f0e0629af6b18a6defea90c7323ed65e).

The really cool thing using this solution is that you can change pitch and volume at any time and in the middle of the sample.

ReikaKalseki:
Do you think the sound looping functionality could be made a standalone library mod, akin to AnimationAPI? I think it would be very useful, and I may use it for my own sounds (falling back to my current system if not present) if it were.

cm0x4D:
> Do you think the sound looping functionality could be made a standalone library mod, akin to AnimationAPI? I think it would be very useful, and I may use it for my own sounds (falling back to my current system if not present) if it were.

Well the whole thing is about 30 lines of code in 2 classes. The Minecraft API is not much more complicated (ITickableSound), so I do not see a need for a dedicated mod or library. Additionally the current implementation is based on EA's transparent node's.

Feel free to copy or use our code as inspiration. "Permission given" ;-)

Domi1993:
I'm testing the release from /feature/turbinesound
But only the generators are play sound when i used them, the gasturbine stays silent.

I built the mod on monday the sounds should be added at this time, is the gasturbine sound not finished yet, or is the mod i've built the problem?

bowserjr2424:
Yeah sound on steam turbine is pretty messed up, I go through a turbine room and it goes in to ear rape mode.

Baughn:
Thanks to cm0x, the sounds are now unencumbered. @ReikaKalseki, I hope you don't mind if we merge rather than squash -- the individual commits can be instructive, and rebasing to remove yours entirely would be a bit difficult.

This is otherwise ready for review and merge.

src/main/java/mods/eln/mechanical/SimpleShaft.kt
```diff
@@ -67,6 +71,24 @@ open class ShaftRender(entity: TransparentNodeEntity, desc: TransparentNodeDescr
     var connectionType: CableRenderType? = null
     open val cableRender: CableRenderDescriptor? = null
     var cableRefresh = true
+    // Sound:
+    private val soundLooper: ShaftSoundLooper?
```
cm0x4D, line 75: Inner ShaftSoundLooper could be replaced by object?

src/main/java/mods/eln/transparentnode/turbine/TurbineSlowProcess.kt
```diff
@@ -0,0 +1,24 @@
+package mods.eln.transparentnode.turbine
```
cm0x4D, line 1: This class is not needed anymore by using the new audio looping API, I have it removed in my PR (#602) already, so it is pointless to convert it to Kotlin.

cm0x4D (approved):
LGTM

ReikaKalseki:
> @ReikaKalseki, I hope you don't mind if we merge rather than squash -- the individual commits can be instructive, and rebasing to remove yours entirely would be a bit difficult.

?

Domi1993:
@ReikaKalseki I think he means that this branch is merged with your sound effects, because it's to difficult to remove the sounds from Rotarycraft, because everyone can download and build this Mod. (I'm not sure if this 100% correct!)

cm0x4D:
@ReikaKalseki What Baughn means is that your sample files will still be in our git history, though never in a release, as they were replaced by other samples.

src/main/java/mods/eln/mechanical/SimpleShaft.kt
Baughn: No, because it's only optionally initialized. I don't create it at all for shaft blocks that don't have sound.

cm0x4D: Did not see that, sorry. Then it's certainly ok.

src/main/java/mods/eln/transparentnode/turbine/TurbineSlowProcess.kt
Baughn: Removed.

Baughn: Un-removed. Attempting to redo history in that way just makes a mess of the git merge.

Much easier if whoever merges to develop last removes it then.

ReikaKalseki:
> @ReikaKalseki What Baughn means is that your sample files will still be in our git history, though never in a release, as they were replaced by other samples.

That is fine.

Issue #579, opened 2016-12-29
[QOL] Right click to install component


OmegaHaxors:
I noticed that street lamps can have wires and bulbs can be installed in them by right clicking while holding relevant component. This functionality is very nice and needs to be implemented in various machines which can accept it:

Installing drills into the auto miner
Refilling pipes in the auto miner
Installing Ferromagnetic Cores
Placing regulators in machines
Upgrading panels with the Solar Tracker

It should not be used for things such as putting materials into machines, for example putting coal into a macerator as that would dramatically increase the instances of accidental inserts.

cm0x4D:
The lamps were some sort of test for that functionality. If we want to adapt this to a broader set of blocks, I would prefer to have a generic way of doing that in one of the base classes. Should imo be not to hard to achieve.

cm0x4D:
Changes are merged, will be part of the next release. Whenever you find another that would make sense, feel free to reopen the issue.

PR #584, opened 2016-12-30
Auto-inserting items to inventory

cm0x4D:
Sometimes it makes actually sense to auto-install items into nodes (#579) just by right-clicking with the correct item.

This PR tries to generalize the concept and adds this feature to a maximum of possible EA elements.

Ready to merge when

- [x] Replace LampSocket implementation by generic process
- [x] Make the process more flexible (more conditions when to actually insert/replace) 
- Add for more items (ideas welcome)
  - [x] Installing drills into the auto miner
  - [x] Refilling pipes in the auto miner
  - ~~Installing Ferromagnetic Cores~~
  - [x] Placing boosters in machines
  - [x] Upgrading panels with the Solar Tracker
  - [x] Entity filter in Turret 
  - ...

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/584)


AMIDIBOSS (changes requested):
Items already in inventory slot may get back into player's inventory

cm0x4D:
> Items already in inventory slot may get back into player's inventory

It is not that simple - what if the itemstack had more than one item? What if the inventory does not have any free slots left? The MC API is not that straightforward either. 

I decided to just drop the existing item and the player can take the item if he wants.

src/main/java/mods/eln/transparentnode/autominer/AutoMinerElement.java
```diff
@@ -30,8 +33,12 @@
 
 public class AutoMinerElement extends TransparentNodeElement  {
 
-	TransparentNodeElementInventory inventory = new TransparentNodeElementInventory(AutoMinerContainer.inventorySize, 64, this);
-	
+	AutoAcceptInventoryProxy accpetingInventory =
```
Baughn: This is misspelled. (IDEA conveniently does spell-checking if you ask it do.)

cm0x4D:
ready to merge

Baughn:
Tested on Erisia, and works fine (so far).

cm0x4D:
Ok, everything that makes sense for me is done. This might be already too much, I fear that we might confuse players with some of those...

Issue #161, opened 2014-12-10
Suggestion: Transistors


ghost:
They would be similar to real transistors... they could adjust between 0 and 1, not just 0 or 1 like a relay, and they would be able to update MUCH faster than a relay. Adding transistors should allow for voltage regulators to be built as well (which I really need).

serathius:
Transistors are little more complicated :)
If you wont to build voltage regulator use H bridge.

ghost:
I wanted it for control purposes really...

MFernflower:
I was also wondering if MOSFETS and TRIACS would be a possibility but I don't know how effective a TRIAC would be in this current simulation system.

PR #585, opened 2016-12-30
Feature/transistors

cm0x4D:
Very simplified and linearized transistor models (#161).

- Models
  - [x] TO92 sixnode case without cooling possibility
  - [ ] TO220 transparent case with cooling possibility
- Behavior
  - [ ] Simple N & P FETs (simplified, linear model)
  - [ ] Semiconductor Relay
  - ... (ideas welcome)

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/585)


AMIDIBOSS:
Idea: Field effect transistor

cm0x4D:
> Idea: Field effect transistor

That is what FET stands for ;-)

AMIDIBOSS:
Oh, i didn't see FET listed. But P and N is a little bit strange to me... Are bipolar transistors there too? 
And do you need any models?

cm0x4D:
@AMIDIBOSS There exist N-channel and P-channel FETs (MOSFETS), depending the polarity of the control voltage. N-channel are used to close the circuit to ground, P-channel rather between the load and the power supply.

JFET and bipolar transistors are probably too hard to simulate and their electrical behavior would not match well a signal voltage as input. So they are not subject to be integrated.

I would need a 1 block model of a TO-220 case for high power transistors and semiconductor devices. The cooling connection should be on the block's side in order to connect dissipators.

cm0x4D:
How could i forget those: linear voltage regulators!!

Lets call then LM7850 and LM78200 😆

https://en.m.wikipedia.org/wiki/78xx

AMIDIBOSS:
I'd recomend to stick to SMPS. But they can still be regulators. 
Or maybe add something like radiator multiblock😆

cm0x4D:
The idea behind the LM's is that they are in a TO220 case that has to be cooled. The more energy they destroy, the hotter they get. The TO220 case can be connected to dissipators. That is the fun about them.

Switching supplies are mostly more complex and would need external components. But why not an advanced DC/DC converter that includes a regulation of the output voltage?

AMIDIBOSS:
Let's make varriable PSU! I'll give model and gui

cm0x4D:
@AMIDIBOSS Thanks for the TO220 case. Maybe the dissipator should be a little higher to avoid odd rendering when a heat cable is connected (Should be at least the height of that cable.

I can to the change when I integrate the code for the model, are you ok with this?

AMIDIBOSS:
You can change whatever you want, i'm ok with that

PR #594, opened 2017-01-08
Changed some Sounds

Domi1993:
Changed Sound of Standard-Alarm to this https://www.youtube.com/watch?v=aqdMHUfRhW0

Add Motor noise to the Waterturbine soundeffect (it's more realistic)

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/594)


cm0x4D:
The alarm sound is nice, the water turbine sound does not loop very nice (have not tested it in-game, I only listened to ogg file), but when using the new sample play API, the looping will be seamless, so the short interruption at the end will be audible and probably very annoying. I will see if I can trimm the loop...

Booth are stereo, which imho does not make that much sense when playing them using OpenAL, so it would make sense to convert them to mono (64kbit/s) ogg files.

Domi1993:
Ok i can convert the sounds to mono

Domi1993:
Finally i had the time to convert the sound files to Mono.
I think the Loops should be as well (water turbine).

Baughn:
The water turbine is not in fact fine, but I think I fixed it. I'll merge this, then overwrite with the fixed one.

Baughn:
For the curious, the fixed-up water turbine (on repeat): https://brage.info/water_turbine_repeat.ogg

PR #602, opened 2017-01-13
Client side audio looping

cm0x4D:
Move audio looping to client side using the new API:

- [x] Turbines
- [x] DC/DC Converter (Transformer)
- [x] All machines
- [x] Autominer
- [x] Fuel Generator
- [x] ~~Lamp Socket~~ (Was just old code needed to be removed)
- [x] Firedetector
- [x] ~~Grid transformer~~ (Not used yet)
- [x] Eletricalpole

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/602)


AMIDIBOSS:
Lamp socket? I guess, ergonomic light bulbs, but... They only Click on startup. That has to be looped? And electrical pole? Does it make any sound at all?

AMIDIBOSS (approved):

cm0x4D:
Theoretically it is done for the electrical pole, I could not test it yet, as it always did explode before I had the chance to hear something...

Baughn:
The electrical pole is actually pretty tough to make explode...

Tie one pole to an electrical source at 3180V, another to a source at 3200V. This should put it at about 85% load.

AMIDIBOSS:
Is there any diffrence between putting 3180 - 3200
And 0 - 20?

AMIDIBOSS:
And what sound will grid transformer output?

cm0x4D:
> And what sound will grid transformer output?

Humming when overloaded like the regular one.

cm0x4D:
> Tie one pole to an electrical source at 3180V, another to a source at 3200V. This should put it at about 85% load.

Thanks @Baughn, that did the trick. What I did not understood is that the load factor actually was 3.3 with a deltaU=20V, so rather 330% rather than 85%.

cm0x4D:
Tried to rebase, it is a complete mess, so #617 replaces this PR.

PR #605, opened 2017-01-14
Feature/grid converter

AMIDIBOSS:
Created big DC/DC converter model to use with powerlines (No it's not a transformer)
So Baughn doesn't have to :)

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/605)


Baughn:
Thanks!

AMIDIBOSS:
No problem!

PR #610, opened 2017-01-20
added pneumaticraft gasoline, assuming it's petrol and not gasohol.

Castone22:
adding pneumaticraft gasoline to the fuelmap. 

This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/610)

### DIFF
--- a/src/main/java/mods/eln/fluid/FuelRegistry.kt
+++ b/src/main/java/mods/eln/fluid/FuelRegistry.kt
@@ -27,6 +27,7 @@ object FuelRegistry {
             "rc ethanol" to 21172000.0, // RotaryCraft, density = 0,79 kg/l, heating value = 26.8 MJ/kg
             "biofuel" to 17826480.0,    // Minefactory Reloaded, (Bioethanol) density = 0.786 kg/l, heating value = 22,68 MJ/l
             "bioethanol" to 17826480.0, // Forestry
+            "gasoline" to 25820000.0,   // PneumaticCraft, density = 0.755 kg/L, heat value = 34,2 MJ/l
             "kerosene" to 34800000.0,   // PneumaticCraft, heat value = 34,8 MJ/l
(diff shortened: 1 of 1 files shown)


src/main/java/mods/eln/fluid/FuelRegistry.kt
```diff
@@ -27,6 +27,7 @@ object FuelRegistry {
             "rc ethanol" to 21172000.0, // RotaryCraft, density = 0,79 kg/l, heating value = 26.8 MJ/kg
             "biofuel" to 17826480.0,    // Minefactory Reloaded, (Bioethanol) density = 0.786 kg/l, heating value = 22,68 MJ/l
             "bioethanol" to 17826480.0, // Forestry
+            "gasoline" to 25820000.0,   // PneumaticCraft, density = 0.755 kg/L, heat value = 34,2 MJ/l
```
cm0x4D, line 30: How could be gasoline not on the gasoline list? ^^

Should fuel and gasoline not have about the same heating value?

cm0x4D (changes requested):

src/main/java/mods/eln/fluid/FuelRegistry.kt
Baughn: As a matter of game-balance, probably not. Buildcraft fuel is the most expensive type on the list, more or less -- in terms of time spent getting it, not machinery. It makes sense for it to have higher density, as that also translates to higher maximum power output.

As a matter of realism... it's easy to see where the gasoline numbers come from, and they make sense. I'm not sure what material 'fuel' is supposed to be.

cm0x4D: Ok then, so fuel is better and close to kerosine. Works.

cm0x4D (approved):

cm0x4D (approved):

cm0x4D:
Merged, thank you @Castone22!

PR #613, opened 2017-01-22
Added Transmission Tower Model

AMIDIBOSS:
^See above^

That's what i done.
Well, It's not my work,
ADK made the model, i just make pull request. He doesn't have Github account.
Ah, the changes:
Added massive transmission tower. It's 10 blocks tall, and 9 blocks wide. Taller than most trees. Will dominate the landscape like real one does. That's what mod needs. It's not steel, i know, it's concrete. But he was bored, and so he made the model.
Hope you like it, or maybe not, we'll make another one, no problem.
We got loOOots of time now.
Now we're onto make that transistor model. Maybe tomorrow.

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/613)


Baughn:
Coool~~~

I wanted something like this. I need to get off my back and do some coding, eh? :)

PR #615, opened 2017-01-24
[Feature Request] Radio tower

AMIDIBOSS:
Okay, here's the idea: Radio tower takes every signal in range of 128 blocks and extends it to 512 blocks. It sucks tons of power, propably powered by 800V. Maybe some gui to select channel? It's basicly a HUGE  signal repeater. Shall not be hard.
I aolso provide you with a model. I think @Baughn asked for it. Not sure. Anyway. It is BIG, not big. TALL, that's the right word.
BTW It's our first model 5 people worked on at the same time.  It took us like 5 min! Crazy isn't it? This way we can make like 15 models a day!
And if you don't like my idea, i won't cry if you use it for something else, i guess.
PS. Looks like the Bigger the model is, the more we enjoy making it!

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/615)


cm0x4D:
Thanks. Just a hint, the rendering for big models is a little complicated, as during rendering it has the same light-level as the main block. So the bigger the models the more unrealistic they may end up when rendered.

AMIDIBOSS:
So, what? Oh, you mean splitting model into smaller chunks, right?

cm0x4D:
No. Not using more blocks as necessary.

cm0x4D:
Just had a look at the model, it looks awesome.

AMIDIBOSS:
Maybe the radio tower should be lit... What do you think?

PR #617, opened 2017-01-25
Client side audio looping

cm0x4D:
replaces #602

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/617)


cm0x4D:
@Baughn can you have a look? I cherry picked and merged manually, hope I did not replace code...

cm0x4D:
@Baughn I added a slew limiter for the generator audio and made it a little less loud, ok with that?

AMIDIBOSS (approved):

Baughn:
I'm fine with that, of course. Usually am.

Maybe the slew limiter should be global, though? (I.e. applied in the looped-sound class, for all such sounds.)

Baughn:
Give it a day or so before merging, though; I'm about to test this branch on Erisia.

Baughn:
Or... less, I guess?

It works, and sounds awesome. I haven't tested absolutely everything, but I've tested most of these. The glitching is entirely gone.

cm0x4D:
> Maybe the slew limiter should be global, though? (I.e. applied in the looped-sound class, for all such sounds.)

Oh yes, good idea. When you are happy and it did not crash anymore, you can merge it. I can add the global slew limit after the merge too.

The generator sample is not yes as smooth as I would like it to be. But for the moment this should do the trick...

PR #620, opened 2017-01-27
PID chip GUI slider step is 1.0.

cm0x4D:
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/620)


Baughn:
I wonder if it could be made logarithmic?

cm0x4D:
Good idea, would actually make sense for the PID parameters. I do not see a technical problem, but first I would really like to integrate the fridge (it is about time...).

Baughn:
Fridge, of all things...

What does it *do*?

cm0x4D:
Ups, was still logged as the bot...

AMIDIBOSS:
Bot came to life😱

PR #626, opened 2017-01-31
Add a config parameter for plate conversion ratio, and set it to 1 by default

Baughn:
Why 1?

It's what literally every other mod uses as the conversion ratio for plate machines. Using anything else just drives people away from using this one.

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/626)


cm0x4D (approved):
LGTM

AMIDIBOSS (commented):
Looks good

PR #627, opened 2017-01-31
Sample and hold chip

cm0x4D:
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/627)

- [x] A recipe would be nice (for all analog chips).


src/main/java/mods/eln/Eln.java
```diff
@@ -5971,7 +6032,7 @@ void recipeThermalIsolator() {
 
     }
 
-    void recipeRegulatorItem() {
+        void recipeRegulatorItem() {
```
Baughn: This looks like an indentation error. Pass it through autoformat?

Baughn:
One comment, otherwise LGTM.

PR #631, opened 2017-02-01
Rewrite UUID manager to be asymptotically optimal

Baughn:
...hopefully!

This got tricksy. Please read carefully.

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/631)


Baughn:
Right, this doesn't work. At present it somehow causes sounds to be duplicated.

Baughn:
Hmm, this patch might not be at fault. I had the same problem happen without it.

Baughn:
In fact this is what @cm0x4D's compressor sound patch was about. Now I get it!

Sounds still sometimes get lost with this patch applied, so far only during times of bad packet loss, so it's probably got to do with networking somehow. Minecraft uses TCP, but I *think* some packets count as optional? And if anything's worth dropping, it'd be these. Might be just WAI.

But I'm not sure you got that right... the compressor sounds odd now.

Might be just unfamiliarity. It's properly synced with the animation.

Well, the fade-in doesn't work so well with coal plates to diamond.

cm0x4D (approved):

cm0x4D:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/electrical-age/electricalage/631)*

PR #633, opened 2017-02-03
Use a no-tool-required material for exporters. Fixes #628.

### LINKED ISSUE #628: [Bug] Converters instantly break with invalid tools
Attempting to use any tool which cannot pick up the converter will cause it to be destroyed instantly anyway, causing you to lose the object.

Baughn:
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/633)

### DIFF
--- a/src/main/java/mods/eln/node/simple/SimpleNodeBlock.java
+++ b/src/main/java/mods/eln/node/simple/SimpleNodeBlock.java
@@ -13,7 +13,6 @@ import net.minecraft.world.World;
 
 public abstract class SimpleNodeBlock extends BlockContainer {
 
-
     protected SimpleNodeBlock(Material material) {
         super(material);
     }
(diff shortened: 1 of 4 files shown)


AMIDIBOSS:
Isn't that "the lazy way"?

Baughn:
Maybe? I don't know. Constructing custom materials is disallowed by the permission set on that class, so I have to assume this is what you're meant to do.

bloxgate:
No, you should be able to make new materials fairly easily.

Edit: Something like:
```java
public ComputerProbeBlock() {
		super(new Material(MapColor.stoneColor) {
			@Override
			public boolean isToolNotRequired() {
				return true;
			}
		});
	}
```

Baughn:
By subclassing and overriding the methods that read its member variables, which will then disagree with the method return values, potentially causing conflicts with anything that happens to read the fields instead?

I mean, they're private, but...

That's still one of the ugliest pieces of code I've seen.

bloxgate:
Good point, though it turns out the requiresNoTool variable defaults to true anyway, so just making a new material of whatever color you want would work too. It'd be nice to have a material class that can have those variables edited though. I wonder if there's already a library for that sort of thing.

PR #636, opened 2017-02-04
Fixes #635

### LINKED ISSUE #635: Devices lose energy
Любые устройства работающие от батареек теряют энергию если их поместить в механизм правым кликом.
Google:
Any device running on batteries lose energy if placed in the right-click mechanism.

Тут понятнее (Надеюсь русский интерфейс не помешает):
Google:
There are clearer (I Hope Russian interface will not interfere):
https://youtu.be/wQtOYnd_zdw

cm0x4D:
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/636)

### DIFF
--- a/src/main/java/mods/eln/node/AutoAcceptInventoryProxy.kt
+++ b/src/main/java/mods/eln/node/AutoAcceptInventoryProxy.kt
@@ -2,6 +2,7 @@ package mods.eln.node
 
 import mods.eln.generic.GenericItemBlockUsingDamageDescriptor
 import mods.eln.generic.GenericItemUsingDamageDescriptor
+import mods.eln.item.electricalitem.BatteryItem
 import net.minecraft.inventory.IInventory
 import net.minecraft.item.ItemStack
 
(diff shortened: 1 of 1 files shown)


src/main/java/mods/eln/node/AutoAcceptInventoryProxy.kt
```diff
@@ -26,8 +27,10 @@ class AutoAcceptInventoryProxy(val inventory: IInventory) {
             if (inventory.getStackInSlot(index) == null) {
                 GenericItemUsingDamageDescriptor.getDescriptor(itemStack)?.let { desc ->
                     if (acceptedItems.any { it.isAssignableFrom(desc.javaClass) }) {
+                        val newItemStack = desc.newItemStack()
+                        (desc as? BatteryItem)?.let { it.setEnergy(newItemStack, it.getEnergy(itemStack)) }
```
Baughn, line 31: If you use apply instead of let, you get this instead of it. (And can usually skip the this.)

That's a matter of taste. This is fine.

PR #637, opened 2017-02-05
Feature/T2 poles

### LINKED ISSUE #546: Grid to Grid Voltage Step-Down Problem / Grid transformer remake Sugesstion
This is not a bug or any  lite problem, it shows deeper in gameplay

How do I step down the voltage not having to make things like this:

Beceuse it's unpractical. I have to step voltage:
Down from 3200 to 800 - to pull it down from grid
Down from 800 to 200 - actual step down
Up from 200 to 800 - to put it back on grid
Can i make a little less silly way?
We only have Grid to ground with transformer
Mabe you should add back again:
Grid to Grid transformer - same polemount, exept without a downlink
Grid to ground transformerless - A downlink, with pole model, but without transformer

I can even provide you with models, if you need/wish

Baughn:
Fixes #546, among others.

This contains two distinct models for T2 poles, both capable of 50kV but one having half the resistance (and taking twice the cable) as the other. Additionally, that grid transformer will finally be back.

Additionally, as a convenience feature (and because the transmission towers have a range of 96 blocks), Eln now searches the player's inventory for HV cable when connecting. 

- [X] Transmission Tower
- [x] Large Transmission Tower
- [ ] Downsize the transmission tower a tad...
- [x] Grid transformer
- [ ] Meet-in-the-middle rendering
- [x] Fix all the rendering glitches I just introduced
- [ ] Tooltips

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/637)


Baughn:
http://i.imgur.com/7aQA1jn.png

...yep, large towers are large.

AMIDIBOSS:
"Eln now searches the player's inventory for HV cable when connecting."  And that's what I wanted to see! And ADK's model looks even bigger in world, but hey! That's a transmission tower!

AMIDIBOSS:
And will downlink be back? 
Have the idea: maybe we shall add grid relay? High voltage, high current. And requires power to open/close I even have an idea for model!

cleverca22:
what about a form of arc flash that can damage players if you try to switch things off the wrong way?

https://www.youtube.com/watch?v=rdQR7-Ap6YQ

AMIDIBOSS:
A bit drastic, but would be cool

Baughn:
This currently works. It's not done yet, but it should be playable if anyone wants to try it.

pooing1234:
@AMIDIBOSS Any news on industrial fans?

KSSilence:
Hi!, Is there somewhere i can download a build with these T2 Poles and transformers in it, that will be compatible with my 1.7.10 server and clients?  I'd like to test this out. I've already got the standard release build on the server, can i just upgrade straight to this, or is that liable to break my existing power grid system thats being built?

AMIDIBOSS:
Unfortunately, you'd have to build it on your own. Or wait a couple hours till I come back.

KSSilence:
Im quite happy to wait,  Building from source is something i'm not really capable of as i have never done it before.

jrddunbr:
It's a pretty popular request, I may write some step by step instructions
on the wiki sometime.

sent from Jared's phone

AMIDIBOSS:
Gimme sec.

AMIDIBOSS:
First try running it on client before you swap server one for it.
And I take no responsibilities...
[ElectricalAge-1.14.2.jar.zip](https://github.com/Electrical-Age/ElectricalAge/files/1745437/ElectricalAge-1.14.2.jar.zip)

KSSilence:
Thank you!

i'l just backup the entire server and use a copy of it to test the mod, the server lives on my NAS at home so its a 1 minute thing to create a backup of it.

KSSilence:
Alright, looks like a direct upgrade from previous version isn't a smart idea, (in my case, on a COPY of my server, so the original is untouched, i aint stupid.), in this case the Gas Turbine APU generator wouldn't self sustain when the load relay closed when it works fine on the release version, so i opened the isolator switch manually to allow it to spin up then closed the switch......    KAPOW!   200KV into the 200v lighting system, and cables at 200K degrees celsius....   the entire building went up in a cloud of explosions. Also a lot of EA items are marked as "NAN<units>" on the hud when you look at them and anything with shafts is missing said shafts... that only happened once the explosions began.

image

screen shot 2018-02-22 at 7 51 27 pm

image

image


It's also pegging the thread that minecraft uses on the server at ~100% lol, which basically never happens with the release version (sits around 60% usually because of the amount of EA stuff i have in one small area).

Any way i can find out what went wrong and report it?
i could always start a fresh map if you want and see if the errors are reproducible, it may be linked to me doing an in place upgrade, or it may be a persistent issue of some sort.

KSSilence:
Alright, Back in action now, though had to replace all the Turbines and Generators (and had to remove ALL of the units in a genset then replacement, if you replace one at a time the replacements bug out).

Generators run now, however the cables still report "NAN" temperature, though they seem to work otherwise without issue.

image

KSSilence:
Everything seems to still be happy now, main generator is online without issues, and i have a 65kw test load at the end of a long run of transmission lines going while i test them out. At 65KW Load, im using about 60% engine throttle once stable, of that 65Kw load, about 8.5Kw is "Internal Loads" for Powerstation lighting and APU Starting battery charging, such internal loads are supplied via Auxiliary Generators connected to the shaft going thru the wall in the previous photo.

In regards to "Downsize the transmission tower a tad..." I think they are perfect how they are actually, i quite like their size, its useful for suspending cables over uneven terrain at maximum distances between poles. 

What would be nice is to go even higher on the voltage, up to ~215-220KV Nominal by going through a second Grid DC-DC converter (i run my generators to regulate at around 3.35-3.4kv. 3.2kv i have found wont provide enough voltage to sufficiently charge 8x 50v batteries in series on the end of a 4:1 DC-DC Converter, but ~3.5KV does. I tried double converting and as i predicted, nothing can tollerate 220kv, or anything close to it, but i was curious.

The slightly higher voltage i run also means i can run long runs of lights without them dimming.

One thing we absolutely need is some proper High Voltage "Pole Mounted Switches" (like those fused cutouts or automatic reclosers you see in the real world), Bonus points if said switches can be controlled via signal cables like Reclosers are. 

It is not practical to have to step everything down to ~3.4Kv, and run like 8+DC-DC pole to Relay to DC-DC Pole setups in parallel just to be able to handle some 80-100kw (Because my Power station can put 90kw to grid easilly) of switching power at a substation. So yeah we need some actual proper transmission voltage rated switches, both for the utility poles and for transmission poles.

Auxiliary "Starter Generators" on the Power take off shaft:
screen shot 2018-02-24 at 12 45 52 am

AMIDIBOSS:
Did U made use of T2 already?

KSSilence:
Yes, thats one of the first things i did once i fixed the bugged out generators.   I've got a ~56.5Kw Test load on the end of quite a long run of those Transmission towers. So far those work without issue. Two Gripes thus far, 1. connecting cables to the grid transformer is a pain because you have to click on a specific side of either the HV end or the LV end, and only on that side will it work rather than anywhere on the correct "end", and gripe 2 is that due to the base design of the transmission tower if you want to get the "quick stats" where you hover over and object and look at it with the cross hairs, you need to be underneath the tower looking up at the middle block of the base, otherwise you can only stat check with a multi-meter.

AMIDIBOSS:
1. Transformer does that to prevent you from screwing up, thus blowing up entire grid.
2.That's work in progress.

PR #638, opened 2017-02-05
Lowpass filter chip

cm0x4D:
- [ ] ~~Use unity gain for Kaiser window~~
- [ ] ~~Find parameters to reduce oscillating on rectangular signals~~
- [x] Icon
- [x] Description
- [x] Save data to NBT
- [x] Logarithmic slider for frequency
- [x] Damn I always forget the recipe 🙄 

All other filter types might not make that much of a sense with simulation frequencies that low.

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/638)


AMIDIBOSS (approved):

Baughn:
On the face of it this looks good, but it's complicated. I'm going to try it on Erisia for a day or two, see if I can find a good use case and test it properly.

Baughn:
They seem to work: https://imgur.com/FVHNjfB.png

src/main/java/mods/eln/gui/GuiHelper.java
```diff
@@ -75,6 +75,13 @@ public GuiVerticalProgressBar newGuiVerticalProgressBar(int x, int y, int width,
         return o;
     }
 
+    public GuiVerticalCustomValuesBar newGuiVerticalCustomValuesBar(int x, int y, int width, int height, Float[] values) {
```
Baughn, line 78: I'll be so happy when we can replace the GUI code with LibLib.

src/main/java/mods/eln/sixnode/AnalogChips.kt
```diff
@@ -685,3 +683,121 @@ class SampleAndHold : AnalogFunction() {
...
+    private var output = 0.0
+
+    override fun process(inputs: Array<Double?>, deltaTime: Double): Double {
+        output = Utils.limit(output + ((inputs[0] ?: 0.0) - output) * feedback * deltaTime, 0.0, 50.0)
+        return output
+    }
+
+    override fun readFromNBT(nbt: NBTTagCompound?, str: String?) {
+        nbt?.apply {
+            feedback = getDouble("feedback")
```
Baughn, line 702: It's very unlikely that you'll need the precision, and floats are smaller. This matters when persisting, or especially for network transfer.

src/main/java/mods/eln/sixnode/AnalogChips.kt
```diff
@@ -685,3 +683,121 @@ class SampleAndHold : AnalogFunction() {
...
+            setDouble("output", output)
+        }
+    }
+}
+
+class FilterElement(node: SixNode, side: Direction, sixNodeDescriptor: SixNodeDescriptor) :
+    AnalogChipElement(node, side, sixNodeDescriptor) {
+
+    companion object {
+        val CutOffFrequencyChangedEvents = 1
```
Baughn: If this is meant to be an enum, then use an enum. See https://kotlinlang.org/docs/reference/enum-classes.html

src/main/java/mods/eln/gui/GuiHelper.java
cm0x4D: Do not know that lib, where can I find infos about?

src/main/java/mods/eln/sixnode/AnalogChips.kt
cm0x4D: Everything on the simulation side is double. I convert it to/from float when serializing to/from client.

src/main/java/mods/eln/sixnode/AnalogChips.kt
cm0x4D: Imho enums are completely broken in Kotlin the same way they are in Java.

cm0x4D: Kotlin 1.1 adds the method `enumValueOf<T>()` which would make enums less a pain to use.

src/main/java/mods/eln/gui/GuiHelper.java
Baughn: There's a limited amount of documentation, unfortunately. I've been browsing the code a bit, and talking to other devs. But: https://github.com/TeamWizardry/LibrarianLib/wiki/GUI-API-Home

1.10 only thus far. There is an 1.7 version, but it's different enough that it isn't worth it.

src/main/java/mods/eln/sixnode/AnalogChips.kt
Baughn: Mm. Well, the 1.1 RC seems stable enough that you can switch to it if you like.

cm0x4D: Oh, I just saw 1.1 is actually out ^^

AMIDIBOSS (approved):

cm0x4D:
@Baughn is there still something open with this PR or can I merge?

Baughn (approved):
Oops, guess I didn't approve.

You can merge. But note that the "Character.valueOf" calls are redundant; you can use plain old characters.

Issue #642, opened 2017-02-07
bug infinite explosions


bjorguv:
VERSION 1.13
after breaking a few cables connected to a creative voltage thing. It started exploding infinitely until I shut the server off. It spammed my console with this message.

http://pastebin.com/du2e9cwS

bjorguv:
Is there a temporary fix for this because I really like your mod and would love to use it on my server but every time this happens it forces me to create a new map.

cm0x4D:
Why a new map? Can you not load the same map again?

bjorguv:
@cm0x4D I can but the explosions are infinite like I said. so when I load the world again the console is spammed VERY rapidly with that message I showed in pastebin. When I toggle explosions in the config file I am still spammed in the console.

cm0x4D:
Ok. This is strange. Normally after restarting the server or reloading the world, the problem is gone. Do you have a map with that problem? Could you send that map? This would really help to track the issue down.

bjorguv:
@cm0x4D Oh, I never actually tried to do either of those things. Sorry about that I misspoke in my response. I will try it now and get back to you.

Also do you know what causes this to happen? If so could you tell me so that I can avoid causing it in the future

cm0x4D:
Well it is a bug where a voltage watchdog is not destroyed along with the block when triggered. This happens very rarely, this is why we could not fix the bug yet.

bjorguv:
@cm0x4D Interesting it happened to me 4 times during a 3 hour period. any idea what could make it more frequent?

cm0x4D:
Do you remember what you exactly did in order to have this problem? 

I have found a problem with a race condition between the GUI and the simulation, but this means that there had to be items with a GUI and not just cables...

bjorguv:
@cm0x4D Oh yes, I seem to have left that part out. The only thing each case had in common was an Electrical Source and cable.

# GUIs INVOLVED

**in the first 2 occurences of the bug**
• _Experimental Transporter GUI_
• _Electrical source (creative energy block) GUI_

**last 2 occurrences**
• _Electrical source (creative energy block) GUI_
• _Lamp supply GUI_

cm0x4D:
Should be fixed in 1.14.0 - Feel free to reopen if it still happens.

PR #644, opened 2017-02-08
Fixes #565 and probably #642

### LINKED ISSUE #565: Failed transformer creates dead zone
Attempting to place a HV to MV transformer will cause the transformer to drop when inserting the 4 MV wires to step down. As a result, literally anything you place in that square will become deleted. This includes things such as flowing water, piston heads, and any block you place.

To make matters stranger, any electrical component placed on the dead block will operate for exactly 1 tick before being deleted and dropped as an item.

This effect will eventually dissipate, presumably since some error checking deletes the fake object.

cm0x4D:
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/644)

### DIFF
--- a/src/main/java/mods/eln/node/NodeBase.java
+++ b/src/main/java/mods/eln/node/NodeBase.java
@@ -153,6 +153,10 @@ public abstract class NodeBase {
 
     boolean destructed = false;
 
+    public boolean isDestructing() {
+        return destructed;
+    }
+
     public void physicalSelfDestruction(float explosionStrength) {
         if (destructed == true) return;
         destructed = true;
@@ -193,7 +197,7 @@ public abstract class NodeBase {
 
     // leaf
     public void onBreakBlock() {
-
+        destructed = true;
         disconnect();
         NodeManager.instance.removeNode(this);
         Utils.println("Node::onBreakBlock()");
--- a/src/main/java/mods/eln/node/six/SixNodeElement.java
+++ b/src/main/java/mods/eln/node/six/SixNodeElement.java
@@ -98,6 +98,9 @@ public abstract class SixNodeElement implements GhostObserver, IPlayer, IPublish
     }
 
     public void connectJob() {
+        // If we are about to destruct ourselves, do not add any elements to the simulation anymore.
+        if (sixNode != null && sixNode.isDestructing()) return;
+
         Eln.simulator.addAllElectricalComponent(electricalComponentList);
         Eln.simulator.addAllThermalConnection(thermalConnectionList);
 
--- a/src/main/java/mods/eln/node/six/SixNodeElementInventory.java
+++ b/src/main/java/mods/eln/node/six/SixNodeElementInventory.java
@@ -137,7 +137,7 @@ public class SixNodeElementInventory implements IInventory, INBTTReady {
 
     @Override
     public void markDirty() {
-        if (null != sixNodeElement) {
+        if (sixNodeElement != null && !sixNodeElement.sixNode.isDestructing()) {
             sixNodeElement.inventoryChanged();
         }
     }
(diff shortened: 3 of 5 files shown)


src/main/java/mods/eln/node/NodeBase.java
```diff
@@ -153,6 +153,10 @@ public NodeBase() {
 
     boolean destructed = false;
```
Baughn, line 154: I'm a bit leery of this sort of patch, so let me think out loud.

It'll *work*, no doubt about that, but only so long as a bunch of special cases are caught. What's actually going on is... the node is getting reconnected after it's been supposedly destroyed, when the UI closes. A symptom of the simulation existing in a parallel world from Minecraft, I suppose; the block is gone, but the node is not.

Ideally we'd tie them a bit more closely together. How often is connectJob called? Often enough... checking for the block's existence every time it's run might degrade performance, or have other problems.

Maybe we could add a process that, at a relatively slow pace, checks each node to make sure it still exists in-world? That would also catch other causes of this sort of thing. It should cry havoc to the console if it finds anything, of course, but it's a fallback and debugging mechanism.

Baughn (approved):
This looks like it'll work, though.

src/main/java/mods/eln/node/NodeBase.java
cm0x4D: > It'll work, no doubt about that, but only so long as a bunch of special cases are caught. What's actually going on is... the node is getting reconnected after it's been supposedly destroyed, when the UI closes. A symptom of the simulation existing in a parallel world from Minecraft, I suppose; the block is gone, but the node is not.

The parallel existence is not optional. It is required as nodes which's blocks are not actually loaded have to be simulated too. The *destructed* attribute was already present before, imho this is the least intrusive way to fix the issue by just using it as how it was actually intended.

> Ideally we'd tie them a bit more closely together. How often is connectJob called? Often enough... checking for the block's existence every time it's run might degrade performance, or have other problems.

This method is only called when there is an interaction of the user with the block or one of it's neighbors or a block gets destroyed. I would not call this often. Further the performance impact of reading this boolean should be not significant at all.

I can remove the check in the connectJob() method, it is just a second test that might catch (or not) other issues, the inventory issue should be already fixed in the respective Inventory classes. But for the actual costs I think it is worth to keep it.

> Maybe we could add a process that, at a relatively slow pace, checks each node to make sure it still exists in-world? That would also catch other causes of this sort of thing. It should cry havoc to the console if it finds anything, of course, but it's a fallback and debugging mechanism.

Imho maybe additionally having such a test in debug mode would make sense to find problems, but I think the impact to the performance could be significant, as this check might force Minecraft to load offline chunks just to do the check.

Baughn:
Oh, I didn't mean that reading the boolean would be expensive. It's reading the Minecraft world to check if the block still exists that would be expensive.

As to scanning... Sure, but nothing is going to change in an unloaded chunk. It'd work well enough even if it skips all unloaded blocks.

cm0x4D:
> Oh, I didn't mean that reading the boolean would be expensive. It's reading the Minecraft world to check if the block still exists that would be expensive.

NodeBase is not part of the Minecraft world, it is part of the EA parallel world. So there is no reading from the Minecraft world at all. 

This list of nodes (NodeManager) seem to work pretty well. I never found a node in the manager that was not destroyed when the block was. Even with the GUI bug, the block ends up destroyed. 

The problem was that by adding the IProcess objects to the simulator (in the connectJob() method), they continue to exist even after  the node instance was destroyed, as the simulator's list retains them.

> As to scanning... Sure, but nothing is going to change in an unloaded chunk. It'd work well enough even if it skips all unloaded blocks.

Ok, then it would make sense. Maybe we could add a setting to turn it on or off?

Baughn:
I'll add the scanner as a follow-up.

PR #674, opened 2017-02-22
Bugfix/save of doom

### LINKED ISSUE #673: Placed blocks and items disappear on world reload.
Using ELN 1.14.1, Forge 1614, and minecraft 1.7.10. Whenever I exit the world and reload, all Electrical Age items that have been placed in the world disappear. My game doesn't crash, but I can't figure out what's causing this.

====

*Admin edit*:

**DO NOT USE 1.14.0 and 1.14.1 on Windows anymore - both probably will break your world.**

**HERE SOME IMPORTANT INFO HOW YOU MIGHT SAVE YOUR MAP:**
For all having this problem, you might eventually be able to restore your world, but you need to manually copy files before starting Minecraft for the next time (using the fixed EA build):

- **If you have this problem, avoid loading the affected saves before you have copied the files!**
- Find the Minecraft save files (location depends platform)
- If there are files with the names electricalAgeWorld0.dat.tmp, electricalAgeWorld1.dat.tmp and * electricalAgeWorld-1.dat.tmp*, create a copy of those in another folder or even better copy the whole folder as a backup.
- Rename the files to electricalAgeWorld0.dat, electricalAgeWorld1.dat and * electricalAgeWorld-1.dat* (Just remove the .tmp suffix). If such files were already there, delete them in order to be able to rename the tmp files.
- Now start Minecraft and load the map. There is a chance that the blocks appear now in your world again.

(Sorry I edited your post, but this is the simplest way to inform everyone having the same issue)

Baughn:
Switch savegame manipulation to the nio API, which should be more atomic. Or, on Windows, at least less likely to fail silently. This also changes the build to require Java 1.7, but that was already the case for Forge so...

Hopefully fixes #673.

---
This change is [Reviewable](https://reviewable.io/reviews/electrical-age/electricalage/674)

### DIFF
--- a/src/main/java/mods/eln/server/ServerEventListener.java
+++ b/src/main/java/mods/eln/server/ServerEventListener.java
@@ -19,9 +19,8 @@ import net.minecraftforge.event.world.WorldEvent.Load;
 import net.minecraftforge.event.world.WorldEvent.Save;
 import net.minecraftforge.event.world.WorldEvent.Unload;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
+import java.io.*;
+import java.nio.file.*;
 import java.util.HashSet;
 import java.util.LinkedList;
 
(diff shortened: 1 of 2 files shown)


src/main/java/mods/eln/server/ServerEventListener.java
```diff
@@ -120,32 +121,31 @@ public void onWorldSave(Save e) {
...
             ex.printStackTrace();
         }
+    }
 
-        //ElnWorldStorage storage = ElnWorldStorage.forWorld(e.world);
-        int idx = 0;
-        idx++;
+    private void replaceFile(Path from, Path to) throws IOException {
+        try {
+            Files.move(from, to, StandardCopyOption.ATOMIC_MOVE);
```
cm0x4D: Files.move(from, to, StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.ATOMIC_MOVE);

I would pass both parameters for the first try.

src/main/java/mods/eln/server/ServerEventListener.java
```diff
@@ -172,4 +172,16 @@ public static void writeToEaWorldNBT(NBTTagCompound nbt, int dim) {
 
     }
 
+    private class FileNames {
+        final Path worldSave;
+        final Path tempSave;
+        final Path backupSave;
+
+        FileNames(WorldEvent e) {
+            String saveName = getEaWorldSaveName(e.world);
```
cm0x4D, line 178: The method getEAWorldSaveName() could be part of this class, it is only used here.

src/main/java/mods/eln/server/ServerEventListener.java
Baughn: I'm pretty sure it won't help, but -- might as well.

cm0x4D: Just read the doc. It is not clear:

> The move is performed as an atomic file system operation and all other options are ignored. If the target file exists then it is implementation specific if the existing file is replaced or this method fails by throwing an IOException.

Again implementation specific 🙄. If we take phrase one, then you are right, REPLACE_EXISTING is not needed.

cm0x4D (approved):

src/main/java/mods/eln/server/ServerEventListener.java
Baughn: Done. Also spotted a bit of a bug while I was doing.

src/main/java/mods/eln/server/ServerEventListener.java
Baughn: It certainly can't *hurt*, so yes. My suspicion is just that it's mapped to rename() on unix and nothing elsewhere.

AMIDIBOSS:
So 14.2 is out?

PR #677, opened 2017-02-24
Fixed ArrayIndexOutOfBoundsException when using /eln command without arguments

cm0x4D:


Baughn (approved):

AMIDIBOSS:
There was an /eln command?

PR #689, opened 2017-03-03
Feature/scanner

Baughn:
This acts basically like a redstone comparator, except when it doesn't. See the tooltip in Scanner.kt.


AMIDIBOSS:
Did you just... 
Remove my model?

Baughn:
Um, no?

I moved it.

src/main/java/mods/eln/sixnode/ElectricalFuse.kt
```diff
@@ -205,11 +205,7 @@ class ElectricalFuseHolderElement(sixNode: SixNode, side: Direction, descriptor:
     }
 
     override fun onBlockActivated(entityPlayer: EntityPlayer?, side: Direction?, vx: Float, vy: Float, vz: Float): Boolean {
-        if (Utils.isPlayerUsingWrench(entityPlayer)) {
-            front = front.nextClockwise
-            sixNode.reconnect()
-            return true
-        }
+        if (super.onBlockActivated(entityPlayer, side, vx, vy, vz)) return true
```
cm0x4D: Maybe 
```java
if (onBlockActivatedRotate(entityPlayer)) return true
```
would be more clear when actually reading the code (multiple locations).

cm0x4D (commented):
Tested it and it works fine ^^.

-The texture is a bit boring. Maybe the side that actually does the scanning should have a different color.

- Maybe a mode switch could be added to the model in order to see without WAILA or the multimeter which mode is active.

src/main/java/mods/eln/sixnode/ElectricalFuse.kt
Baughn: Concur.

AMIDIBOSS:
@cm0x4D Good Idea

cm0x4D (approved):

AMIDIBOSS (commented):
And what that export script is for?

cm0x4D:
Attention: sixnode models and transparentnode models do not have the same settings when exported to wavefront files. You may need to check and adapt the rendering code. Imho it would be great to have one setting for all models, even if this means we have to adapt the rendering code.

Baughn:
The transparentnode models *also* don't have all the same settings, because I didn't find it documented anywhere. Yeah, long-term that's the plan. I'll probably do it around the time we're porting this to 1.10, since then I'll need to redo a lot of the rendering anyway.

For now, I just didn't want to lose the script.

Baughn:
...and also, there are some optimizations I tend to do on models; e.g. joining every object that can be joined, so that only a single draw call is needed. If I can move that into the *script*... which I can, of course.

It'll be great. You'll see. ;)

Baughn:
I think this is ready now. Any final comments?

cm0x4D:
On my machine it crashes on startup: http://pastebin.com/usFYvcxB

I suspect the run of the blender script as a problem...

Baughn:
Interesting.

I've un-run it. How do things look now?

cm0x4D:
My GIT client is now in nirvana. Cannot do anything anymore...

cm0x4D:
That was not funny. I had to merge in order to stash in order to checkout another branch in order to delete and re-create and checkout this branch. On the bright side, now it does not crash anymore 😄

cm0x4D:
Everything works for me now.

@AMIDIBOSS: Could you add indicators for the modes next to the LEDs, would make it clearer.

PR #690, opened 2017-03-05
Case for DC-DC converter

cm0x4D:
- [x] Add case item
  - [x] Item (code)
  - [x] Icon
  - [x] Recipe
- [x] Add case item slot to DC/DC converter GUI
- [x] Adapt rendering code to use case only in the presence of the item
- [x] Add animation for casing door.


src/main/java/mods/eln/Eln.java
```diff
@@ -6645,6 +6647,12 @@ void recipeMiscItem() {
             Character.valueOf('g'), new ItemStack(Blocks.glass_pane),
             Character.valueOf('c'), new ItemStack(Items.dye, 1, 1));
 
+        addRecipe(findItemStack("Casing"),
+            "ppp",
```
Baughn: This is a bit expensive for what's a cosmetic item. Maybe make the recipe output 4 or 8 items?

src/main/java/mods/eln/Eln.java
```diff
@@ -6645,6 +6647,12 @@ void recipeMiscItem() {
             Character.valueOf('g'), new ItemStack(Blocks.glass_pane),
             Character.valueOf('c'), new ItemStack(Items.dye, 1, 1));
 
+        addRecipe(findItemStack("Casing"),
```
Baughn: This is a bit expensive for what's a cosmetic item. Maybe make the recipe output 4 or 8 items?

src/main/java/mods/eln/misc/Obj3D.java
```diff
@@ -518,7 +518,9 @@ public boolean loadFile(final String filePath, boolean reload) {
                 String line;
                 while ((line = bufferedReader.readLine()) != null) {
                     String[] words = line.split(" ");
-                    if (words[0].equals("o")) {
+                    if (words[0].startsWith("#")) {
+                        // # is a comment - ignore line.
```
Baughn: Unknown verbs are already ignored. I guess we should log a debug warning for those?

Baughn (approved):

src/main/java/mods/eln/Eln.java
cm0x4D: Good point. Totally agree.

PR #692, opened 2017-03-07
Feature/emergency lamp

cm0x4D:
- [x] Balance (charge power, battery capacity, consumption)
- [x] Icon
- [x] Recipe
- [x] Allow to charge via lamp supply
- [x] Add cable render when connected using cables
- [x] Add VoltageWatchdog


AMIDIBOSS (approved):

AMIDIBOSS (approved):

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
```diff
@@ -65,6 +70,14 @@ class EmergencyLampDescriptor(name: String, val cable: ElectricalCableDescriptor
     override fun getFrontFromPlace(side: Direction?, player: EntityPlayer?)
         = super.getFrontFromPlace(side, player).inverse()
 
+    override fun canBePlacedOnSide(player: EntityPlayer?, side: Direction?): Boolean {
+        if (side == Direction.YN) {
```
Baughn: Why not?

I mean, it might look silly. But that's on the player, and denying the option isn't realistic, is it?

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
```diff
@@ -167,6 +169,8 @@ class EmergencyLampElement(sixNode: SixNode, side: Direction, descriptor: SixNod
         electricalComponentList.add(chargingResistor)
         slowProcessList.add(process)
         slowProcessList.add(NodePeriodicPublishProcess(sixNode, 2.0, 0.5))
+        slowProcessList.add(VoltageStateWatchDog().set(load).setUNominal(desc.cable.electricalNominalVoltage)
```
AMIDIBOSS: Why it doesn't feature Max. Current/Power?

AMIDIBOSS:
Does it feature emergency light cube?

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
cm0x4D: The resistance is constant for each model and taking I = U/R and P = U^2/R into account we see that only voltage is actually significant in this case.

cm0x4D:
What is an emergency light cube?

AMIDIBOSS:
@cm0x4D You might haven't seen one before, I't a quite new device
It's a cube-shaped device hidden in celling. It monitors if there is a grid power. If there's not, it turns on it's powerLED emitting small amounts of light, but still enough to safely get out of the building in case of emergency. And if there is a grid power, it lights up green led to show it's working alright. In emergency mode (as manufacturer claims) they shall last up to 48 hours. Because "They use Li-Po".
And if cube needs to be serviced, it lights up red.

cm0x4D:
Oh, these emergency lamps do exactly that.

src/main/java/mods/eln/node/IPublishable.kt
```diff
@@ -10,7 +10,8 @@ interface IPublishable {
 class published<T>(var value: T) : ReadWriteProperty<IPublishable, T> {
     override fun getValue(thisRef: IPublishable, property: KProperty<*>): T = this.value
     override fun setValue(thisRef: IPublishable, property: KProperty<*>, value: T) {
+        val changed = this.value != value
```
Baughn: This is correct... but I initially thought it was incorrect. It's worth reading https://kotlinlang.org/docs/reference/equality.html, as it's a difference from Java.

(Reference equality could give false negatives here.)

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
```diff
@@ -0,0 +1,320 @@
...
+    init {
+        voltageLevelColor = VoltageLevelColor.fromCable(cable)
+        changeDefaultIcon("emergencylamp")
+    }
+
+    fun draw(onCeiling: Boolean = false, on: Boolean = false, mirrorSign: Boolean = false) {
+        if (onCeiling) {
+            mainCeiling.draw()
+
+            if (on)
```
Baughn: Please use braces any time the if clause is more than a single line. Or just always; I like to follow https://google.github.io/styleguide/javaguide.html#s4.1-braces on this.

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
```diff
@@ -0,0 +1,320 @@
...
+    enum class Event(val value: Byte) {
+        TOGGLE_POWERED_BY_CABLE(1),
+        SET_CHANNEL(2)
+    }
+
+    val desc = descriptor as EmergencyLampDescriptor
+    val load = NbtElectricalLoad("load")
+    val chargingResistor = ResistorSwitch("chargingResistor", load, null)
+    var on = false
+        set(value) {
```
Baughn: It should be possible to use IPublishable here. Add an extra, optional lambda parameter for "more-stuff-to-do-post-setting"?

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
```diff
@@ -0,0 +1,320 @@
...
+            if (field != value) needPublish()
+            field = value
+            sixNode.lightValue = if (value) desc.lightLevel else 0
+        }
+    var charge = desc.batteryCapacity / 2
+    var poweredByCable = false
+        set(value) {
```
Baughn: Ditto... though rather than immediately adding pre-change and post-change parameters, reconnection is common enough that a 'reconnect' optional bool might be appropriate.

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
```diff
@@ -0,0 +1,320 @@
...
+
+            LampSupplyElement.channelMap[channel]?.forEach {
+                val distance = it.element.sixNode.coordonate.trueDistanceTo(sixNode.coordonate).toFloat()
+                if (distance < closestDistance && distance <= it.element.range) {
+                    closestDistance = distance
+                    closestPowerSupply = it
+                }
+            }
+
+            if (closestPowerSupply != null && closestPowerSupply!!.element.getChannelState(closestPowerSupply!!.id)) {
```
Baughn, line 122: You shouldn't need the !!s, given the explicit null test. (And IDEA should be telling you that.)

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
```diff
@@ -0,0 +1,320 @@
...
+            if (charge > 0) {
+                on = true
+                charge = Math.max(charge - desc.consumption * deltaT, 0.0)
+            } else {
+                on = false
+            }
+        }
+    }
+
+    override fun initialize() {
```
Baughn, line 151: Honestly, this could all be put in init{}. Doesn't make a difference, I suppose.

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
```diff
@@ -0,0 +1,320 @@
...
+class EmergencyLampGui(private var render: EmergencyLampRender)
+    : GuiScreenEln() {
+    private var buttonSupplyType: GuiButton? = null
+    private var channel: GuiTextFieldEln? = null
+    private var charge: GuiVerticalProgressBar? = null
+
+    override fun initGui() {
+        super.initGui()
+        buttonSupplyType = newGuiButton(18, 12, 140, "")
+        channel = newGuiTextField(19, 38, 138)
```
Baughn, line 267: I don't think this should be able to return null.

Baughn (changes requested):

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
cm0x4D: Ok. Let them do silly things. It will not hurt anyone.

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
cm0x4D: Good idea.

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
cm0x4D: It is strange, IDEA tells me exactly the opposite. It does not compile without the non-null assertion.

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
cm0x4D: Actually it makes sense, as the SixNode and TransparentNode classes actually initialize some parameters before the method initialize() is called. For example the front is already set and a disconnect of the other nodes has been executed before.

src/main/java/mods/eln/sixnode/EmergencyLamp.kt
cm0x4D: Ok. lateinit is definitely the better solution here.

Baughn (approved):

PR #702, opened 2017-03-13
Better DC/DC case

cm0x4D:


cm0x4D:
@AMIDIBOSS: Please check if you are ok with the changes. 1 door felt a little awkward, so I added a second door and I gave the texture a little more life and dept.

PR #703, opened 2017-03-15
Feature/modern street light

cm0x4D:
It is basically working, but...
- [x] Texture needs some optimization
- [ ] Move light source to the effective position of the light
- [ ] Add recipe
- [ ] Add 2D icon to inventory
- [ ] Fix the bug that makes the ghostblock's direction static


AMIDIBOSS:
Wait with that! I still need to add 3 more models
Double lamp, rightangle lamp and quad lamp. 
BTW is it doable? And do you want that?

AMIDIBOSS (approved):

cm0x4D:
@AMIDIBOSS Yeah, why not. I will prepare the blender file, so you can add those variants as different objects in the same blender file and I will just render the right one in code. Ihmo that is better than adding 4 different models. Ok?

AMIDIBOSS:
OK

PR #712, opened 2017-03-24
Let users the choice between symbols and renderings of the blocks

cm0x4D:
Not ready to merge, a lot of icons have to be added...

- [x] Chips
- [x] Thermistor
- [x] Switches
- [x] Signalsource
- [x] Scanner
- [x] Rheostat
- [x] Powerresistor
- [x] Source
- [x] Breaker
- [x] Power transmission antennas
- [x] Button
- [x] Batteries
- [x] 2x3 Solar panels 
- [x] Replace actual rendered blocks and items by textures 
- [x] Remove use2DIcon() method for items


AMIDIBOSS:
Nice feature!
˙˙˙sƃuıɹǝpuǝɹ ǝɥʇ ǝsooɥɔ p,ı

cm0x4D:
@Baughn This PR does not only add the config to choose between symbols and renderings, it does replace every rendered item with an icon, so this might actually help in porting to 1.10+...

src/main/java/mods/eln/generic/GenericItemBlockUsingDamageDescriptor.java
```diff
@@ -26,12 +27,19 @@ public GenericItemBlockUsingDamageDescriptor(String name) {
     }
 
     public GenericItemBlockUsingDamageDescriptor(String name, String iconName) {
-        this.iconName = "eln:" + iconName.replaceAll(" ", "").toLowerCase();
+        changeDefaultIcon(iconName);
```
Baughn: Might be better named "setDefaultIcon".

src/main/java/mods/eln/gridnode/GridDescriptor.java
```diff
@@ -62,13 +62,17 @@ public void draw(float idealRenderingAngle) {
 
     @Override
     public void renderItem(ItemRenderType type, ItemStack item, Object... data) {
-        objItemScale(obj);
```
Baughn, line 65: I converted this whole thing to Kotlin in a separate PR I haven't finished yet...

Oh, I won't stop you. But this will be fun.

src/main/java/mods/eln/sixnode/Scanner.kt
```diff
@@ -28,6 +28,10 @@ class ScannerDescriptor(name: String, obj: Obj3D) : SixNodeDescriptor(name, Scan
     val main = obj.getPart("main")!!
     val leds = arrayOf("LED_0", "LED_1").map { obj.getPart(it) }.requireNoNulls()
 
+    init {
+        voltageLevelColor = VoltageLevelColor.SignalVoltage
```
Baughn, line 32: These shouldn't normally change, so an abstract / open val in the superclass might do better.

src/main/java/mods/eln/sixnode/TreeResinCollector/TreeResinCollectorDescriptor.java
```diff
@@ -52,7 +51,7 @@ void draw(float factor) {
 
     @Override
     public boolean use2DIcon() {
```
Baughn: Are there any use2DIcon implementations left that return false?

Baughn (changes requested):

src/main/java/mods/eln/misc/VoltageLevelColor.java
```diff
@@ -14,6 +14,7 @@
     MediumVoltage("medium"),
     HighVoltage("high"),
     VeryHighVoltage("veryhigh"),
+    Grid("grid"),
```
AMIDIBOSS, line 17: There will be a few grid voltages, 2 for now. What colors will they have?

src/main/java/mods/eln/sixnode/TreeResinCollector/TreeResinCollectorDescriptor.java
cm0x4D: I can remove it already for GenericItemBlockUsingDamageDescriptor subclasses.

Some items (GenericItemUsingDamageDescriptor subclasses) do still return false in order to do some custom rendering. Have to address them later.

src/main/java/mods/eln/sixnode/Scanner.kt
cm0x4D: It is just a field in the base class, I would have to add a getter and setter to convert it to a property?

src/main/java/mods/eln/misc/VoltageLevelColor.java
cm0x4D: Same as high and very high, but with a pattern.

src/main/java/mods/eln/sixnode/Scanner.kt
Baughn: No, you can override fields in Kotlin. ..though then it'd need to be Kotlin. Perhaps not.

Baughn (approved):

PR #718, opened 2017-04-08
 Improvement tr_TR 

anar4732:


cm0x4D:
@anar4732 Thanks for your contribution. We use [Transifex](https://www.transifex.com/electrical-age/eln/dashboard/) now to translate the mod into other languages. 

I will try to include your changes there, but best is you add translations in the future there (It is much more convenient anyway...)

cm0x4D:
I integrated the changes on Transifex. Thanks for your contribution!

Issue #693, opened 2017-03-08
Startup crash with AOBD


mpevner:
Using the builds from Safranil, 453 [0] and higher crash on startup with AOBD present.
The pertinent line appears as follows:
```
cpw.mods.fml.common.LoaderException: java.lang.IllegalAccessError: tried to access field mods.eln.Eln.maceratorRecipes from class ganymedes01.aobd.recipes.modules.ElectricalAge
```
And the full log is here: https://paste.ee/p/UN7FZ

I've tossed this as an issue on the AOBD repo [1], but thought to also bring it to your attention, as AOBD has not been worked on in some time.

[0] http://ci.safranil.fr/job/ElectricalAge/453/
[1] https://github.com/ganymedes01/Another-One-Bites-the-Dust/issues/129

Baughn:
Can you try removing 'private' from maceratorRecipes?

mpevner:
I adjusted as follows in Eln.java and compiled:
```
RecipesList maceratorRecipes = new RecipesList();
void registerMacerator(int id) {
```
It did not work however.

Baughn:
Ok. I'll make sure it gets fixed before the next release; thanks for telling us.

mpevner:
I gave this another go, explicitly changing every Private RecipesList to Public RecipesList in Eln.java. This seems to work so far, but I can see this as being a rather hacky fix. On a probably related note, I think this breaks NEI. I am investigating further in that regard.

mpevner:
Amendment: This was an issue with the mod NEI Integration, which was using an S instead of a Z. I presume at some point the spelling in this mod changed. A custom compile of NEI Integration later and it is resolved.
...I'm not entirely sure of my reading this, but the issue for NEI appears to be magnetiser/magnetizer. In this matter NEI crashes on attempt to view any recipe.
Error excerpt as follows:
```
java.lang.NoSuchFieldError: magnetiserRecipes
	at tonius.neiintegration.mods.electricalage.RecipeHandlerMagnetizer.getRecipes(RecipeHandlerMagnetizer.java:21)
	at tonius.neiintegration.mods.electricalage.RecipeHandlerElnBase.loadCraftingRecipes(RecipeHandlerElnBase.java:71)
	at codechicken.nei.recipe.TemplateRecipeHandler.loadCraftingRecipes(TemplateRecipeHandler.java:331)
[...]

PR #724, opened 2017-04-17
Shade kotlin into ELN's libs Package

### LINKED ISSUE #722: Incompatability with certain mods in a modpack

bloxgate:
Closes #722 

This prevents conflicts when another mod with a different version of Kotlin is present.

It would probably be better to close 722 by simply not including Kotlin at all and requiring Forgelin as a dependency, but this would also work for the time being.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -56,6 +56,8 @@ minecraft {
     // TODO(Baughn): After moving to 1.10, this should be replaced with the shadow plugin,
     // e.g. as used in https://github.com/Emberwalker/Laundarray/blob/master/build.gradle
     srgExtra "PK: org/apache/commons/math3 mods/eln/libs/org/apache/commons/math3"
+    srgExtra "PK: kotlin mods/eln/libs/kotlin"
+    srgExtra "PK: org/jetbrains/annotations mods/eln/libs/annotations"
 
     replaceIn "Version.java"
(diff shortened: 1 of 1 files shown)


PyroTheManiac:
you did not explain how to do this at all

cm0x4D (approved):
Not an expert on forge gradle stuff, but LGTM

Baughn:
Your commit introduces whitespace errors. Please fix.

Also, I assume you've tested this both against a server *with* other Kotlin-dependent mods and *without*?

cm0x4D:
@Baughn I updated the branch `release/1.14` to `1.14.2`, so if we would like to release an update (1.14.3) fixing the Kotlin lib issue, you can just cherry-pick into the `release/1.14` branch and create a new release. I would wait with the 1.15 release for the T2 poles...

bloxgate:
Yes, I've tested this with the same modpack as was used in #722 and on a normal forge server/client. Everything seems to be fine, though it looks like #693 will happen in the pack that @PyroTheManiac is using unless AOBD is disabled for Eln.

PyroTheManiac:
@bloxgate did you do something to make aobd work in your modpack/instance?

bloxgate:
I made a custom build using the changes in this PR, then disabled AOBD support for ELN and everything worked fine.

PR #763, opened 2017-09-05
Move third party APIs to API module

bloxgate:
Moves third party APIs that don't need to be included into the API module.

If I understand how this works, everything in the API module can be accessed by the main module, but it is not included when the jar is built.

My tests seem to show that this is indeed the case.


Baughn:
Merged manually. (You have an extra commit from me that's not supposed to be there.)

PR #765, opened 2017-09-20
implement clamping in the PID chip

lashtear:
* allows higher Ki without a reset pin
* add state inspection for the waila display


Baughn:
The PID function for the signal processor already implements clamping.

Rather than duplicating the code here, can you merge the two? They should both be using the same code.

cm0x4D:
Did this in #832: The PID chip uses the code of the signal processor's equation.

I just discovered that the clamping in the signal processor is done between -1 .. 1 ([here](https://github.com/Electrical-Age/ElectricalAge/blob/3e7db53eac084b4f2770139949630d01f72a8767/src/main/java/mods/eln/solver/Equation.java#L764)), but the actual range of the output is rather 0 .. 1 (0V .. 50V), so I guess the actual clamping in the signal processor is not 100% correct there...

Baughn:
Is this PR still relevant?

lashtear:
The problem is still relevant, @Baughn -- while the signal processor works, the PID-chip itself is terribly susceptible to windup, and refactoring the signal proc functions to be usable in other areas looks like a larger undertaking.  Can look into that I guess.

lashtear:
Closing as undesired / out of scope for 1.7.

PR #798, opened 2017-10-14
fix #797 by returning false

### LINKED ISSUE #797: Infinite loop in SixNode::canPlaceBlockAt
Drop a sixnode-anything around a Rannuncarpus, and the server will freeze. This is our fault.
```
@Override
    public boolean canPlaceBlockAt(World par1World, int par2, int par3, int par4) {
        Utils.println("CACATOTAL1");
        while (true)
            ;// DON'tCALL THAT XD
    }
```

Wat.

lashtear:
This will unfortunately prevent players from rebuilding circuits via
Rannuncarpus or similar.

Other remaining loops still need to be fixed.

Tested solo and with Botania/Baubles deobf.


lashtear:
Travis is failing as a result of prior commits, I think?

Baughn (approved):

PR #804, opened 2017-10-27
Updating Transifex client used in Travis.

metc:
Simplify a bit the logic to install the Transifex client and push the generated language file.
* Use API token set in Travis UI.
* Install the latest transifex Python client and do not compile from source any more. Do this only when needed and not for every build. Should speed-up the build time a bit.

### DIFF
--- a/script/transifex.sh
+++ b/script/transifex.sh
@@ -4,11 +4,15 @@
 #
 # Generate and push the source language file to Transifex.
 # This script is called after every successful build on Travis CI.
-# It relies on $TRANSIFEX_PASSWORD being set in .travis.yml
-# The Transifex client must be installed.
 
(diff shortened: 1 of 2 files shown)


script/transifex.sh
```diff
@@ -4,21 +4,24 @@
 #
 # Generate and push the source language file to Transifex.
 # This script is called after every successful build on Travis CI.
-# It relies on $TRANSIFEX_PASSWORD being set in .travis.yml
-# The Transifex client must be installed.
 
 if [ "$TRAVIS_REPO_SLUG" == "Electrical-Age/ElectricalAge" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$TRAVIS_BRANCH" == "develop" ]
```
metc, line 8: @cm0x4D Just realized this one will be hard to sanity check and get green before merging since the script this is not executed for PR :)

Baughn:
Is this PR still up to date? If so, let's try it.

PR #808, opened 2017-11-04
Fix #806

### LINKED ISSUE #806: Radio signals can be blocked by air blocks
...if they're mod-added, non-default air blocks.

The erronous code is at https://github.com/Electrical-Age/ElectricalAge/blob/develop/src/main/java/mods/eln/sixnode/wirelesssignal/WirelessUtils.java#L181, which should use World.isAirBlock instead of checking the block ID.

lashtear:


Baughn (approved):
LGTM. For the future, you'll want to title PRs so Github picks up on the issue fix.

PR #822, opened 2017-12-16
let transformers hold more wires

lashtear:
This allows stuffing more wires into the transformer block for more complicated ratios and makes a (weak) stab at rendering the results.

screenshot

### DIFF
--- a/src/main/java/mods/eln/transparentnode/transformer/TransformerContainer.java
+++ b/src/main/java/mods/eln/transparentnode/transformer/TransformerContainer.java
@@ -21,10 +21,10 @@ public class TransformerContainer extends BasicContainer {
 
     public TransformerContainer(EntityPlayer player, IInventory inventory) {
         super(player, inventory, new Slot[]{
-            new SixNodeItemSlot(inventory, primaryCableSlotId, 58, 30, 4, new Class[]{ElectricalCableDescriptor.class},
+            new SixNodeItemSlot(inventory, primaryCableSlotId, 58, 30, 64, new Class[]{ElectricalCableDescriptor.class},
                 SlotSkin.medium,
                 new String[]{tr("Electrical cable slot")}),
-            new SixNodeItemSlot(inventory, secondaryCableSlotId, 100, 30, 4,
+            new SixNodeItemSlot(inventory, secondaryCableSlotId, 100, 30, 64,
                 new Class[]{ElectricalCableDescriptor.class}, SlotSkin.medium,
                 new String[]{tr("Electrical cable slot")}),
             new GenericItemUsingDamageSlot(inventory, ferromagneticSlotId, 58 + (100 - 58) / 2, 30, 1,
--- a/src/main/java/mods/eln/transparentnode/transformer/TransformerDescriptor.java
+++ b/src/main/java/mods/eln/transparentnode/transformer/TransformerDescriptor.java
@@ -83,22 +83,22 @@ public class TransformerDescriptor extends TransparentNodeDescriptor {
             fero.draw();
             if (priCableNbr != 0) {
                 GL11.glPushMatrix();
-                float y = (priCableNbr - 1) * 1f / 16f;
-                GL11.glTranslatef(0f, -y, 0f);
+                float separation = 0.5f / (priCableNbr + 1);
+                GL11.glTranslatef (0f, separation-4/16f, 0f);
                 for (int idx = 0; idx < priCableNbr; idx++) {
                     sbire.draw();
-                    GL11.glTranslatef(0f, 2f / 16f, 0f);
+                    GL11.glTranslatef (0f, separation, 0f);
                 }
                 GL11.glPopMatrix();
             }
             if (secCableNbr != 0) {
                 GL11.glPushMatrix();
                 GL11.glRotatef(180, 0f, 1f, 0f);
-                float y = (secCableNbr - 1) * 1f / 16f;
-                GL11.glTranslatef(0f, -y, 0f);
+                float separation = 0.5f / (secCableNbr + 1);
+                GL11.glTranslatef (0f, separation-4/16f, 0f);
                 for (int idx = 0; idx < secCableNbr; idx++) {
                     sbire.draw();
-                    GL11.glTranslatef(0f, 2f / 16f, 0f);
+                    GL11.glTranslatef (0f, separation, 0f);
                 }
                 GL11.glPopMatrix();
             }


AMIDIBOSS:
Great

AMIDIBOSS:
I love it.

AMIDIBOSS:
approved

src/main/java/mods/eln/transparentnode/transformer/TransformerContainer.java
```diff
@@ -21,10 +21,10 @@
 
     public TransformerContainer(EntityPlayer player, IInventory inventory) {
         super(player, inventory, new Slot[]{
-            new SixNodeItemSlot(inventory, primaryCableSlotId, 58, 30, 4, new Class[]{ElectricalCableDescriptor.class},
+            new SixNodeItemSlot(inventory, primaryCableSlotId, 58, 30, 64, new Class[]{ElectricalCableDescriptor.class},
```
AMIDIBOSS, line 24: Isn't 64 like too much?

src/main/java/mods/eln/transparentnode/transformer/TransformerContainer.java
```diff
@@ -21,10 +21,10 @@
...
                 SlotSkin.medium,
                 new String[]{tr("Electrical cable slot")}),
-            new SixNodeItemSlot(inventory, secondaryCableSlotId, 100, 30, 4,
+            new SixNodeItemSlot(inventory, secondaryCableSlotId, 100, 30, 64,
```
AMIDIBOSS, line 27: ?

src/main/java/mods/eln/transparentnode/transformer/TransformerContainer.java
AMIDIBOSS, line 24: Does it "fit" on the core?

AMIDIBOSS (changes requested):

lashtear:
14 is the max for visually distinct rings.  15 they are starting to blend together, and at 16 it's a solid orange bar.  Higher than that and you start getting really ugly z-fighting.  I didn't worry about that too much as we could probably use a better model for the wire or draw it via a scaled block primitive (at least until we redo for 1.10+ models)

I can set the limit to be 14 we think that is a good idea; it has the nice side effect of being less than 16 and thus preventing people from skipping voltage tiers so easily using better transformer ratios...?

src/main/java/mods/eln/transparentnode/transformer/TransformerDescriptor.java
```diff
@@ -83,22 +83,22 @@ void draw(Obj3DPart fero, int priCableNbr, int secCableNbr, boolean hasCasing, f
...
+                    GL11.glTranslatef (0f, separation, 0f);
                 }
                 GL11.glPopMatrix();
             }
             if (secCableNbr != 0) {
                 GL11.glPushMatrix();
                 GL11.glRotatef(180, 0f, 1f, 0f);
-                float y = (secCableNbr - 1) * 1f / 16f;
-                GL11.glTranslatef(0f, -y, 0f);
+                float separation = 0.5f / (secCableNbr + 1);
```
Grissess, line 97: A call to glScalef around here might accomplish two things:

- Thin out the models on the Y axis enough that they remain visually distinct at higher numbers;
- Get rid of this slightly complicated math to calculate a separation--the separation can be counted in the same units as the Y-width (height?) of the model itself :)

On that note, I wouldn't mind the stacking limit as being the limit for a transformer, at least for experimentation. To assuage concerns of tier-skipping, I'd rather introduce a config or make them undesirable at high ratios for some reason (e.g., they become lossy, they have high capacitance, etc.).

Grissess: (NB: I don't know where your transform origin is, and that *might* have an undesirable effect of needing to multiply various translate parameters in this method :)

lashtear: Already tried that; the existing wires (called sbire model in the code) are introduced as a regular object from blender, and pre-translated.  We'd have to un-translate first at and that point it's better to scrap it.

Grissess: I'm amiable to the "scrap it" route, it's definitely simple enough to be practical to draw right in the code :)

AMIDIBOSS:
I thinking 16 is the best solution here.  Skiping steps shouldn't be a problem. Plus some transformer windings look like solid copper if you don't look close enough.

jrddunbr:
I am considering merging this into LLv11 or LLv12. Any thoughts/changes desired?

Baughn:
The rings are geometrically really simple. It should be possible to squash them proportionally to the number of rings, perhaps above a minimum threshold, using simple OpenGL scaling.

Something like this, maybe?

```
glScalef(1.0, 6.0 / Math.max(6.0, ringCount), 1.0)
```

Of course you'll need to preserve the matrix around that call, if that isn't already happening. And this is a rather inefficient way to use OpenGL. That shouldn't matter, unless people plop down hundreds of the things in a single chunk, but do optimize out the call for ringCounts below 7.

Baughn:
We've got this in Erisia now. I'm still waiting for an update PR so I can merge into develop. :)

jrddunbr:
This is in my branch, with updated rendering code that @Grissess wrote, so if/when you pull that in it will be good to go. He worked on the renderer a bit and made it technically support up to 64 cables, but we set the cap to 16.

OmegaHaxors:
Would like to see it go to 64 to both increase the amount of ratios you can do, while also making it more expensive to use in the process. 16 is pretty acceptable right now though, and I've personally been using a lot of 15:16 transformers for small stepups.

lashtear:
You need a rebase and... scaling?  That'll require un-translation, scaling, and re-translation of the ring model, unless someone wants to make a new ring asset.  Or we make one in code (it's just a squashed cube, ugh...)

jrddunbr:
We already made these changes work for LL. It's also in the Erisia build now.

OmegaHaxors:
You think we could have those thick cables back for the smaller ratios? 1:4 looks really silly right now.

lashtear:
If it's merged everywhere everyone wants, shall I close the PR?

jrddunbr:
I'm indifferent. I suspect we're going to be closing a lot of 1.7 tickets
with 1.12+

lashtear:
Closing as obsolete / undesired, can revisit later.

Issue #812, opened 2017-11-08
[Bug] Power Dishes Lose Efficiency In Rain... Underground.


OmegaHaxors:
Normally when it rains the power dishes' efficiency tanks. This is because of increased moisture in the air attenuating the energy and causing it to dissipate as heat. This is all fine and well but when you're deep underground it still behaves like this despite being completely isolated from the harmful effects of the rain.

AMIDIBOSS:
http://digicoll.manoa.hawaii.edu/techreports/PDF/USBM-083.pdf

OmegaHaxors:
As interesting of a read as that is, the humidity in a cave doesn't suddenly go away because it stopped raining. If anything, it should always have lowered efficiency underground. Though it would be nice if there was a proper way to reinforce the dish from the elements.

AMIDIBOSS:
That's it! Humidity stable underground. But how do you check if it's underground?

AMIDIBOSS:
See what i mean: https://sketch.io/render/sk-bb7052249b9a0e44008ff96d46a18f67.jpeg

OmegaHaxors:
Villages have an algo to determine if something is a 'house' i'm sure it can be worked around to determine if an area is an underground.

AMIDIBOSS:
Yes, but villagers algo. includes existance of a door

OmegaHaxors:
Caves now require doors.

AMIDIBOSS:
XD

Baughn:
Simple enough. Minecraft already caches the "is underground" value for every block, to make lighting work.

AMIDIBOSS:
Oh
OK
Guess doable?

OmegaHaxors:
Turns out they will lose efficiency in 'the rain' while being placed in the middle of a desert.

AMIDIBOSS:
Humidity rises even if there's no rain...

PR #825, opened 2017-12-21
Wirelesscache

jrddunbr:
I converted the WirelessUtils.java file to Kotlin, added a getChunk() method to Coordonate, as well as new function parts for the revamped wireless raytrace code.


jrddunbr:
Thanks to @Grissess's effective rewrite of the math portions of the code I started (I'm not great at this kind of math), the virtualDistance calculation appears to work as intended now. It is mathematically working better than before - the new code handles signal traversal through blocks far better, including parts of blocks, which is currently somewhat broken - it could miss or count doubly for blocks previously, depending on the angle. Now, we use the distance that it travels through a block, which is more accurate over longer distances.

The only thing I would check at this point is whether the cache is working as we wanted. I think that it isn't quite what we wanted yet, so we may want to check on that, and make any necessary clarifications. Particularly, I'm looking at [line 40 and 42 in this code](https://github.com/jrddunbr/ElectricalAge/blob/d6ca96334162ffbd9b63520fd81a36317acfcc58/src/main/java/mods/eln/sixnode/wirelesssignal/WirelessUtils.kt#L40) - that may not be quite correct. Other edits may need to take place elsewhere for the loaded/unloaded chunk stuff - but that may be handled by L47 in the same file.

We did discuss making the hardness of the block factor into the equation, but Baughn commented that some of the hardness's are rather wacky (look at stone and planks for example), so that should wait for another time. It would, however, not be difficult to add that feature in, given the code here. Perhaps that and/or #812 could be done at the same time in a future pull request.

cm0x4D:
For WAILA integration we used Google guava's cache (See [here](https://github.com/Electrical-Age/ElectricalAge/blob/3a8c989a6943af0bfc5bc254f138f4f6d5a10d3c/src/main/java/mods/eln/integration/waila/WailaCache.java)) and it worked very well. Might be of use here too...

src/main/java/mods/eln/misc/FloatLinear.kt
```diff
@@ -0,0 +1,130 @@
+package mods.eln.misc
+
+data class Vec3f(var x: Float, var y: Float, var z: Float) {
```
Baughn, line 3: I would really like some comments here. I mean, yes, it's a 3-element vector class with linear operations, but the other classes down below are nowhere near as obvious.

src/main/java/mods/eln/misc/FloatLinear.kt
```diff
@@ -0,0 +1,130 @@
...
+    companion object {
+        fun fromEndpoints(a: Vec3f, b: Vec3f) = Line3f(a, b.sub(a))
+    }
+
+    fun pointAt(u: Float) = origin.add(vector.mul(u))
+    fun midpoint() = pointAt(0.5f)
+    fun opposite() = pointAt(1.0f)
+    fun length() = vector.mag()
+
+    fun allIntersectionsAlongU(): List<Float> {
```
Baughn, line 61: If you got this algorithm from somewhere, please include a link. That goes for all of this.

src/main/java/mods/eln/misc/FloatLinear.kt
```diff
@@ -0,0 +1,130 @@
...
+        val PLANE_Y = Plane3f(Vec3f.AXIS_Y, 0.0f)
+        val PLANE_Z = Plane3f(Vec3f.AXIS_Z, 0.0f)
+        val AXIS_PLANES = arrayOf(PLANE_X, PLANE_Y, PLANE_Z)
+    }
+
+    constructor(normal: Vec3f, point: Vec3f) : this(normal, -normal.dot(point))
+
+    fun pointOn() = normal.mul(-offset)
+
+    fun isctLineU(ln: Line3f): Float? {
```
Baughn, line 113: What does "isctLine" do, exactly? Documentation is good, and better names would also be good.

src/main/java/mods/eln/sixnode/wirelesssignal/WirelessUtils.kt
```diff
@@ -0,0 +1,246 @@
...
+            total += costMap.getOrDefault(chunk, 0.0)
+        }
+
+        cachedCost = total
+        return total
+    }
+}
+
+
+object WirelessUtils {
```
Baughn, line 63: This object probably shouldn't exist; it seems to have no state whatsoever. Functions can exist outside of classes, and can still be private or internal. I'm not saying it's the only way, but give it a bit of thought.

Baughn (changes requested):

jrddunbr:
I'm closing this because it's probably not going to happen for Minecraft 1.17.10. I'll keep the code locally on my GitHub until we're ready for it in the new Eln.

PR #826, opened 2017-12-22
Renamed probes to be more specific

jrddunbr:
I renamed the various probes to Inline Electrical Probe, Electrical Probe, Inline Temperature Probe, Temperature Probe, and then updated the tip strips and inventory images to reflect their names and uses better - since it is unclear previously that you can't use the ```Thermal Probe``` (which I renamed to ```Inline Temperature Probe```) to measure the temperature of a cable, I made the tip strip explain that better. The images were also updated to reflect the new names, C, E, T, V didn't make a lot of sense, and didn't line up with the existing names. I did IE, E, IT, T for Inline Electrical, Electrical, Inline Temperature, and Temperature, respectively. I could have doe IEP, EP, etc, but it would have been a bit cluttered on a small icon, so I decided not to.

I also increased the smaller icons to all be the the same size - 32x32. Some were 16x16, and one was 32x32.

I expect some discussion to take place, I want to see what people think.


jrddunbr:
I was also thinking of changing how the non-inline probes look, but I don't know what would make sense.

I drew this one, but I'm not sure. I also played with thicker edges to the rectangle, and making it square. Making it square causes a strange visual illusion that I don't like. The ticker lines makes it look more epic, I suppose.

temperatureprobe

temperatureprobe

lashtear:
Looks good to me, though I'm not sure what translation work might need to be done for other localizations.  Also, using letters/words in icons is usually a bad idea for exactly that reason.  Not sure they are large enough to have sensible electric/thermal and inline/terminal symbols in there, though.

The probe shape shouldn't be square anyway.

jrddunbr:
I was considering other locales. I can do all of the English based ones that we may have.

jrddunbr:
Did this work previously? I should enumerate it there if so. I have no idea how that ties into the language file, but I guess it does? I've never used another language in Minecraft

https://github.com/Electrical-Age/ElectricalAge/blob/develop/src/main/resources/assets/eln/lang/en_US.lang#L583

src/main/java/mods/eln/sixnode/electricalsensor/ElectricalSensorDescriptor.java
```diff
@@ -60,11 +60,11 @@ public void renderItem(ItemRenderType type, ItemStack item, Object... data) {
...
         super.addInformation(itemStack, entityPlayer, list, par4);
         if (voltageOnly) {
-            list.add(tr("Measures voltage on cables."));
+            list.add(tr("Measures voltage of cables."));
             list.add(tr("Has a signal output."));
         } else {
-            list.add(tr("Measures electrical values on cables."));
-            list.add(tr("Can measure Voltage/Power/Current"));
+            list.add(tr("Measures electrical values of cables."));
+            list.add(tr("Can measure Voltage, Power, or Current"));
```
Baughn, line 67: Nit: Typically you don't want that last comma.

src/main/resources/assets/eln/lang/en_US.lang
```diff
@@ -261,7 +261,7 @@ Street_Light.name=Streetlight
 Suspended_Lamp_Socket.name=Suspended Lamp Socket
 Tachometer.name=Tachometer
 Temperature_Probe.name=Temperature Probe
-Thermal_Probe.name=Thermal Probe
+Thermal_Probe.name=Inline Temperature Probe
```
Baughn, line 264: IIRC this can also measure thermal wattage. If so, then it isn't merely a temperature probe. If not... that would be a useful addition.

src/main/java/mods/eln/sixnode/electricalsensor/ElectricalSensorDescriptor.java
jrddunbr: Ahh, the [Oxford Comma](https://en.wikipedia.org/wiki/Serial_comma) returns :D

src/main/resources/assets/eln/lang/en_US.lang
jrddunbr: It does measure the wattage. Perhaps they should be called ```Thermal Probe``` and ```Inline Thermal Probe```? I'm not really sure. We could also make it more clear in the tip strip that it does thermal watts (which I think it may already do)

cm0x4D: If we are renaming those things, why not rename all of them to ... Sensor? IMHO this word is much more popular than "Probe".

cm0x4D: For the icons I propose another change:
How about to change the inline probe (or sensor) icon so that it is clear that the probe is inline? Like this:
electricalprobe

cm0x4D (dismissed):

src/main/resources/assets/eln/lang/en_US.lang
jrddunbr: Oh, I really like the look of that. Indeed, sensor does sound much better. I will work on that real soon.

Some people are mentioning that we should consolidate the probes to one of each type. I think that for now we push this, and then later I can try consolidating the sensors together - combining them is (for me) significantly harder to do.

cm0x4D: The two types of sensors makes imho a lot of sense:
For the inline version, the connection must be opened in order to place it and a cable has to be placed into the sensor, the other sensor just connects to the side of the cable with the drawback that serial physical values as electrical current (and electrical power, as current is needed to calculate the power) and thermal power can not be measured with that kind of probe.

I do not see a simple way to combining them into one type for each...

Can you revert the icons in your PR so that I can push new versions without conflicts to dev?

jrddunbr: I would say push your icons now, and I will pull out my new icons. I would mention however that I don't think this is finished yet - I still need to fix the translations so that the text is properly translated, from what I saw, it was not doing it right. Perhaps I'm missing some part of the build process - I was told that there's something that does it automatically.

I really dislike the ```sensor``` icons (the non-inline ones), they feel really busy and look like batteries or polarized capacitors. One of them was labeled ```c```, which made even less sense. There's no industry standard icons for that (as far as I'm aware), so it's up to our imagination to put something good there.

PR #832, opened 2018-01-05
PID chip uses signal processor's primitives

cm0x4D:


cm0x4D:
IMHO it does not make sense for the other (trivial) chips to use the signal processor's primitives, so the PR is ready to merge...

Baughn:
Right... I'm just going to merge this, then fix any remaining bugs in the PR myself.

Baughn (approved):

PR #835, opened 2018-01-20
Fix wrong unit when using multimeter on a breaker

nikola3244:


cm0x4D (approved):

cm0x4D:
Thank you for the fix.

PR #843, opened 2018-04-04
Transmitter and Receiver Antenna

setsunafseiei:
I would like to request to add Very High Transmitter and Receiver Antenna
Also requesting a large 3x3 and 5xr5 transmitter and receiver antenna which has built in transformer like the Utility Pole with Transformer and change its name to microwave power antenna.


Baughn:
This is not a sane pull request, and doesn't seem to include the code you're trying to add at all.

Issue #820, opened 2017-12-13
Client-side sound handler taking atrocious amounts of time


Baughn:
E.g. as in https://i.imgur.com/L0gM6oA.png

Are we leaking sound objects?

jrddunbr:
[Commenting this line](https://github.com/Electrical-Age/ElectricalAge/blob/c2eeaa32a7cac8b8ae35499c772e54160a0646a1/src/main/java/mods/eln/sound/LoopedSoundManager.kt#L11) makes the problem go away (*and effectively removes looped sounds*). This narrows it down a little perhaps?

jrddunbr:
I looked around a bit, and I don't think that we are handling sound objects very well. It would appear that when the looped sounds are disposed, it tosses all of them at the same time, and things can be added on every render tick (it would appear, at least for the turbine blocks). I traced `<instance of LoopedSound.kt>.active` around, and it's only called on destruction, and it appears that destruction affects all looped sound objects (and occurs on block breakage).

I'm no expert here.. I could be wrong.

jrddunbr:
```
[16:50:00] [Client thread/INFO] [STDOUT]: [mods.eln.sound.LoopedSoundManager:process:29]: java.lang.IllegalArgumentException: value already present: mods.eln.mechanical.ShaftRender$ShaftSoundLooper@949f7c1
```
I placed a lot of generators down in a test world with nothing else in it. >20
This happens often when you place 30+.

EDIT: L29 is in a different location than normal because I added a few println statements. It happens in the try/catch loop.

PR #865, opened 2018-07-15
fixed sound bug 

jrddunbr:
I fixed the sound bug by setting limits on how far away looped sounds can be heard by the player.
The default is set to 16 blocks.
I also added a configuration option to increase/decrease distance:
```
debug {
    D:maxSoundDistance=16.0
}
```
I have found through rigorous testing that 16 blocks is about how far sounds travel normally. If you place _**a lot**_ of generators in a single area (you really have to be trying), you may need to lower this number to 8 to completely eliminate clipping.

Regardless, there will be no more intense lag caused by the sound engine (like in #820), if the default is left alone, since the sounds are unregistered from the sound manager. (I am pretty sure that the bug was caused by keeping all of the sounds loaded for the whole world, and continuing to register them to play at no volume)

### DIFF
--- a/src/main/java/mods/eln/Eln.java
+++ b/src/main/java/mods/eln/Eln.java
@@ -311,6 +311,8 @@ public class Eln {
     public static boolean noSymbols = false;
     public static boolean noVoltageBackground = false;
 
+    public static double maxSoundDistance = 16;
+
     @EventHandler
     public void preInit(FMLPreInitializationEvent event) {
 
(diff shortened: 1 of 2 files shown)


Baughn:
LGTM, but there's one improvement you should make. This is called a lot of times per second, making performance important, and the square root function is slow as molasses...

You don't need the true distance. Comparing squared distances will give you the same result, while being much faster.

jrddunbr:
Right. Should have looked at the underlying implementation of `player.getDistance`. I'll look into it later today or tomorrow.

jrddunbr:
That should be it.

PR #870, opened 2018-08-12
Microfarad support

### LINKED ISSUE #813: [Tooltip] Capacitance value Rounds to 0F
When putting in lots of dielectrics into a capacitor, the amount of capacitance per redstone drops. Eventually the value from 1 redstone drops so low that the tooltip rounds to 0F making it appear as if there is no capacitance at all. This is not the case as it still holds charge, increasing with redstone.

jrddunbr:
Fixes #813 by supporting µF's

### DIFF
--- a/src/main/java/mods/eln/misc/Utils.java
+++ b/src/main/java/mods/eln/misc/Utils.java
@@ -226,8 +226,10 @@ public class Utils {
 
     public static String plotValue(double value) {
         double valueAbs = Math.abs(value);
-        if (valueAbs < 0.001) {
+        if (valueAbs < 0.0001) {
             return "0";
+        } else if (valueAbs < 0.000999) {
+            return String.format("%1.2fµ",value * 10000);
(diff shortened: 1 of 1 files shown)


Baughn (approved):
LGTM. I'm sure there's a library we could use, but let's save that for later.

jrddunbr:
I'll echo since we talked in chat, this adds support for micro-anything, including (but probably not limited to) microvolt, microamps, microwatts, and microfarads, which means that they can expose previously unknown or somewhat unexpected behavior by the MNA at higher precision (such as switches apparently being 1 gigaohm resistors when open).

Two options - merge as is and see all of this (and perhaps propose fixes for odd behavior) or make it so that it only adds microfarad support and not the others.

Baughn:
Let's merge it as-is. I'm all for exposing bugs, and we can increase the "closed" resistance another few orders of magnitude.

PR #872, opened 2018-08-20
A bunch of small fixes

rbb30:
Fixed a lot of small errors some files had most changes due to forge changing how you track a block/entity position.


src/main/java/mods/eln/sixnode/electricaldatalogger/ElectricalDataLoggerRender.java
```diff
@@ -51,7 +51,7 @@ public void draw() {
                 drawSignalPin(front.inverse(), new float[]{6.37f, 6.37f, 5.67f, 6.12f});
             }
         }
-        descriptor.draw(log, side, front, this.tileEntity.xCoord, this.tileEntity.zCoord);
+        descriptor.draw(log, side, front, this.tileEntity.getPos().getX(), this.tileEntity.getPos().getZ());
```
Baughn, line 54: Please factor out the getPos, so it isn't called more than once. IDEA should make this simple; it has an 'extract variable' option.

src/main/java/mods/eln/sixnode/electricalfiredetector/ElectricalFireDetectorSlowProcess.java
```diff
@@ -69,44 +70,44 @@ public void process(double time) {
             detectionBBCenter.copyFrom(element.getCoordonate());
             switch (element.side) {
                 case XP:
-                    detectionBBCenter.x -= maxRangeHalf;
+                    detectionBBCenter.pos.add(-maxRangeHalf, 0, 0);
```
Baughn, line 73: Just checking, but does pos.add *mutate* the position -- or return a new one?

src/main/java/mods/eln/sixnode/electricalfiredetector/ElectricalFireDetectorSlowProcess.java
```diff
@@ -69,44 +70,44 @@ public void process(double time) {
...
-                            detectionBBCenter.z + dz);
+                        Block block = detectionBBCenter.world().getBlockState(new BlockPos(detectionBBCenter.pos.getX() + dx, detectionBBCenter.pos.getY() + dy,
+                            detectionBBCenter.pos.getZ() + dz)).getBlock();
                         if (block.getClass() == BlockFire.class) {
                             fireDetected = true;
 
                             Coordinate coord = element.getCoordonate();
-                            List<Block> blockList = Utils.traceRay(coord.world(), coord.x + 0.5, coord.y + 0.5, coord.z + 0.5,
-                                detectionBBCenter.x + dx + 0.5, detectionBBCenter.y + dy + 0.5, detectionBBCenter.z + dz + 0.5);
+                            List<Block> blockList = Utils.traceRay(coord.world(), coord.pos.getX() + 0.5, coord.pos.getY() + 0.5, coord.pos.getZ() + 0.5,
```
Baughn, line 106: Ditto factoring.

src/main/java/mods/eln/sixnode/lampsocket/LampSocketProcess.java
```diff
@@ -109,32 +112,26 @@ public void process(double time) {
...
                     for (int idx = 0; idx < lamp.socketDescriptor.range + light; idx++) {
                         // newCoord.move(lamp.side.getInverse());
-                        vp.xCoord += vv.xCoord;
-                        vp.yCoord += vv.yCoord;
-                        vp.zCoord += vv.zCoord;
-
+                        vp.addVector(vv.xCoord, vv.yCoord, vv.zCoord);
                         c.setPosition(vp);
-                        Block b = c.getBlock();
+                        Block b = c.getBlockState().getBlock();
```
Baughn, line 117: Most of the time, where we used to pass Blocks around, we should be passing BlockStates instead. Is this the right thing?

src/main/java/mods/eln/sixnode/lampsocket/LampSocketProcess.java
```diff
@@ -109,32 +112,26 @@ public void process(double time) {
...
                         }
                     }
 
                     if (!exit) {
-                        Block b = c.getBlock();
+                        Block b = c.getBlockState().getBlock();
 
-                        if (b != Blocks.air) {
-                            b.updateTick(c.world(), c.x, c.y, c.z, c.world().rand);
+                        if (b != Blocks.AIR) {
```
Baughn, line 133: This isn't a _new_ bug, but you should use a function that's called something like isBlockAir (IIRC?), rather than checking for it being this specific air block. Virtual light blocks are "air", but not Blocks.AIR, and still shouldn't block processes like this. Ditto Railcraft's heat blocks, etc. etc.

The method is probably on the BlockState rather than the Block, by the way. See my earlier comment.

Baughn (changes requested):

PR #893, opened 2019-02-04
create model for cross dim txrx dish

Castone22:
Create a satellite dish asset per @Baughn for use in ELN to be used for a cross dim transfer system.


Castone22:
image

Baughn (approved):

PR #894, opened 2019-02-05
Cable Quality Update

OmegaHaxors:
After a metric boatload of tedious work, the update is finally ready.
This update adds Shoddy, Quality, and Perfect cables for each tier.

Shoddy cables are cheaper to craft but are only half as good as the full block. They're terrible.
Normal cables are crafted as normal, and are untouched.
Quality cables are crafted with a complicated refining chain, and represent 8 copper plates and have 2x capacity.
Perfect cables have 4x capacity but require a remarkable amount of copper plates be processed heavily and are EXTRAORDINARILY expensive.

The Very High Voltage Quality/Perfect cables are stupid expensive. Perfect for a post-game goal!

EDIT: Also contains a balance update which makes 200V machines run faster and consume more power


OmegaHaxors:
This is my first time doing a pull request to main, let me know if I did it properly

Baughn:
It's not fundamentally a problem, but you will note that the list of commits include a lot of commits that have nothing to do with this one. I'm not going to review them all in one PR, and not all are ready for merging at all -- like my T2 transmission tower stuff.

You can either wait for those respective PRs to get merged, which might take 1-2 months, or you can rebase your changes to not have those dependencies -- I think that should be possible for this set. If you want to do the latter, then make sure to poke me in advance so I can walk you through it without risking data loss.

Baughn:
There also happen to be conflicts with what's in develop right now. Normally the fix for that would be for you to merge develop back into your branch, first, but that's pointless if it's going to be rebased anyway.

OmegaHaxors:
I'm only intending to push the last four commits, is there any way to trim the ones that aren't relevant?

Baughn:
Two ways.

- Rebase; remove the others. This will break anyone who's built on your work.

- First PR the others, so they aren't *new* commits. Unfortunately some of those represent the unfinished transmission tower code, which isn't going to be finished anytime soon.

OmegaHaxors:
I'll close this and send another PR next time with only the commits that i want.

PR #896, opened 2019-02-05
Cable quality update

OmegaHaxors:
Adds Resistive dust, which replaces the function of coal dust in resistors. Crafts from coal dust in a ratio of 1:8

Improved the power draw of 200V machines to 1,000 so they're actually worth the upgrade cost

Added custom registering system for cables which cleans up the code plenty. This also adds 3 new types of Cable Quality: Shoddy with half capacity, Quality with double capacity, and Perfect with 10x the capacity.

Renamed 'normal' cables to Standard cables to be consistant with the naming scheme of the new cables

Complex crafting chain to make Pure Copper, used to craft Quality cables. (Known issue: The bucket and lead is consumed in the process)

Glorious Copper which requires excessive folding of Pure Copper which makes Perfect Cables.

Turrets track faster and cost more to run passively, no longer target dead enemies, remark "gotcha" upon killing an enemy, and generally are better at aiming and killing their targets

Batteries are cheaper to recycle and have realistic battery capacity
Life Batteries have higher resistance so they are no longer a direct upgrade

Added recipe to macerate silicon from flint


OmegaHaxors:
Apologies, I should have tested to see if it compiled before attempting to push the update.
Since you had removed CableRSFactor yet some of my code still used it, I simply removed any reference to it and now things should work properly.

OmegaHaxors:
Did a lot of compatibility changes. It compiles now and I haven't noticed any errors.

src/main/java/mods/eln/Eln.java
```diff
@@ -1189,9 +1212,82 @@ private void registerElectricalCable(int id) {
...
+                LVU * 1.3, LVP() * 1.2,// electricalMaximalVoltage,
+                // electricalMaximalPower,
+                30,// electricalOverVoltageStartPowerLost,
+                cableWarmLimit, -100,// thermalWarmLimit, thermalCoolLimit,
+                cableHeatingTime, cableThermalConductionTao// thermalNominalHeatTime,
+                // thermalConductivityTao
+            );
+            sixNodeItem.addDescriptor(subId + (id << 6), desc);
         }
+        {
```
Castone22: perhaps something like
```java
public class RegistrationHelper {

  public static registerPowerCableTier(string tier, int subId, int voltage, int power, float colorBase){
    name = TRNAME(Type.NONE, "Quality" + tier + "Voltage Cable");
    shoddyRender = new CableRenderDescriptor("eln", "sprites/cableshoddy.png", colorBase, 0.95f);
    shoddyDesc = new ElectricalCableDescriptor(name, shoddyRender, "Not ideal... but it gets the job done.", false);
    shoddyCableDescriptor = desc;

    desc.setPhysicalConstantLikeNormalCable(voltage, power() / 2, 0.10 / 20, // electricalNominalVoltage,
    // electricalNominalPower,
    // electricalNominalPowerDrop,
    LVU * 1.3, LVP() * 1.2,// electricalMaximalVoltage,
     // electricalMaximalPower,
    30,// electricalOverVoltageStartPowerLost,
    cableWarmLimit, -100,// thermalWarmLimit, thermalCoolLimit,
    cableHeatingTime, cableThermalConductionTao// thermalNominalHeatTime,
    // thermalConductivityTao
    );
  sixNodeItem.addDescriptor(subId + (id << 6), desc);
  }
  // doing this for each cable type in the tier, this could be loops further internally
}
```
Then you'd use it like
```
RegistrationHelper.registerCableTier("Low", 6, LVU, LVP(), 1.95f)

**Comment by OmegaHaxors** [comment]
I hope you've liked the changes I did to the wire registry, I worked really hard on it

**Comment by Castone22** [comment]
Given what it was, massive improvement

File: src/main/java/mods/eln/Eln.java [hunk 2]
```diff
@@ -253,13 +255,29 @@
     public int electricalInterSystemOverSampling;
 
     public CopperCableDescriptor copperCableDescriptor;
```
jrddunbr, line 255: Please indent the code properly.

src/main/java/mods/eln/Eln.java
```diff
@@ -630,6 +649,7 @@ public void preInit(FMLPreInitializationEvent event) {
         registerMiningPipe(17);
         registerTreeResinAndRubber(64);
         registerRawCable(65);
+		registerArc(69);
```
jrddunbr: Indent properly, please.

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,135 +1172,133 @@ private void registerElectricalCable(int id) {
...
+        subId = 3;
+        while (cablevoltage <= 3) {
+            renderspriteindex = 0;
+            subId++;
+            name = TR_NAME(Type.NONE, qualitylist.get(cablequality) + voltagelist.get(cablevoltage) + "Voltage Cable");
+
+            renderspriteindex = cablequality;
+
+            switch (cablevoltage) {
+                case 0: power = LVP(); nominalcablevoltage = LVU; renderfloatindex = 0; break;
```
jrddunbr: Newlines needed after every semicolon. You also need a `default` statement.

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,135 +1172,133 @@ private void registerElectricalCable(int id) {
...
+                case 1: power = MVP(); nominalcablevoltage = MVU; renderfloatindex = 2; break;
+                case 2: power = HVP(); nominalcablevoltage = HVU; renderfloatindex = 4; break;
+                case 3: power = VVP(); nominalcablevoltage = VVU; renderfloatindex = 4; renderspriteindex = renderspriteindex + 4; break;
+            }
+            activerendermodel = new CableRenderDescriptor("eln",renderlist.get(renderspriteindex), rendercablesizes.get(renderfloatindex), rendercablesizes.get(renderfloatindex+1));
+            desc = new ElectricalCableDescriptor(name,activerendermodel,"Haxors was Here", false);
+            nominalcablepower = power*(cablequality+1) / 2;
+                switch (cablequality) {
+                    case 0:
+                        switch (cablevoltage) { case 0: LQlowVoltageCableDescriptor = desc; break; case 1: LQmeduimVoltageCableDescriptor = desc; break; case 2: LQhighVoltageCableDescriptor = desc; break; case 3: LQveryHighVoltageCableDescriptor = desc; break;}
```
jrddunbr: These lines are way too long, you need to have newlines after every semicolon. Separate the `case` lines onto their own lines. You are missing a `default:` catch all, might I suggest using the default as the way cables used to be?

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,135 +1172,133 @@ private void registerElectricalCable(int id) {
...
+                    case 1:
+                        switch (cablevoltage) { case 0: lowVoltageCableDescriptor = desc; break; case 1: meduimVoltageCableDescriptor = desc; break; case 2: highVoltageCableDescriptor = desc; break; case 3: veryHighVoltageCableDescriptor = desc; break; }
+                    case 2:
+                        switch (cablevoltage) { case 0: HQlowVoltageCableDescriptor = desc; break; case 1: HQmeduimVoltageCableDescriptor = desc; break; case 2: HQhighVoltageCableDescriptor = desc; break; case 3: HQveryHighVoltageCableDescriptor = desc; break; }
+                    case 3:
+                        switch (cablevoltage) { case 0: PQlowVoltageCableDescriptor = desc; break; case 1: PQmeduimVoltageCableDescriptor = desc; break; case 2: PQhighVoltageCableDescriptor = desc; break; case 3: PQveryHighVoltageCableDescriptor = desc; break; }
 
-            sixNodeItem.addDescriptor(subId + (id << 6), desc);
+                }
+            desc.setPhysicalConstantLikeNormalCable(nominalcablevoltage,nominalcablepower,0.10/20,nominalcablevoltage*1.3,nominalcablepower*1.2,30,cableWarmLimit,-100,cableHeatingTime,cableThermalConductionTao);
```
jrddunbr: This line is pretty long. I would do this:
```
desc.setPhysicalConstantLikeNormalCable(
    nominalcablevoltage,
    nominalcablepower,
    0.10/20,
    nominalcablevoltage*1.3,
    nominalcablepower*1.2,30,
    cableWarmLimit,
    -100,
    cableHeatingTime,
    cableThermalConductionTao
);
```
Perhaps a comment on the end of each one about what on earth it's doing to increase readability.

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,135 +1172,133 @@ private void registerElectricalCable(int id) {
...
 
-            highVoltageCableDescriptor = desc;
+            creativeCableDescriptor = desc;
 
-            desc.setPhysicalConstantLikeNormalCable(HVU, HVP(), 0.025 * 5 / 4 / 20,// electricalNominalVoltage,
-                // electricalNominalPower,
-                // electricalNominalPowerDrop,
-                HVU * 1.3, HVP() * 1.2,// electricalMaximalVoltage,
-                // electricalMaximalPower,
+            desc.setPhysicalConstantLikeNormalCable(VVU, VVU * VVP, 0.025 * 5 / 4 / 20 / 8 / 9001, //what!?
```
jrddunbr: You *actively* made this worse by failing to indent this and by removing the comments, and adding "what" instead. If you're going to question something, please investigate what it does, instead of removing the comments and making it more work for someone to fix later.

src/main/java/mods/eln/Eln.java
```diff
@@ -1429,7 +1451,7 @@ private void registerBattery(int id) {
             name = TR_NAME(Type.NONE, "Life Oriented Battery");
 
             BatteryDescriptor desc = new BatteryDescriptor(name,
-                "BatteryBig", batteryCableDescriptor, 0.5, true, true, voltageFunction, stdU,
+                "BatteryBig", batteryCableDescriptor, 0.5, true, false, voltageFunction, stdU,
```
jrddunbr: Considering I have no idea what changing this from `true` to `false` does clearly, it may be worth commenting this section. Please break it out after each comma on it's own line and add a comment.

src/main/java/mods/eln/Eln.java
```diff
@@ -1465,6 +1487,26 @@ private void registerBattery(int id) {
             transparentNodeItem.addDescriptor(subId + (id << 6), desc);
         }
         {
+            subId = 6;
+            name = TR_NAME(Type.NONE, "Experimental Battery");
+
+            BatteryDescriptor desc = new BatteryDescriptor(name,
+                "BatteryBig", batteryCableDescriptor, 0.5, true, false, voltageFunction, stdU * 2,
```
jrddunbr: Break out and comment like above.

src/main/java/mods/eln/Eln.java
```diff
@@ -2743,7 +2785,7 @@ private void registerTurbine(int id) {
             name = TR_NAME(Type.NONE, "200V Turbine");
             double RsFactor = 0.10;
             double nominalU = MVU;
-            double nominalP = 500 * heatTurbinePowerFactor;
+            double nominalP = 2000 * heatTurbinePowerFactor; // it was 2000 before
```
jrddunbr: and still is?

src/main/java/mods/eln/Eln.java
```diff
@@ -2888,7 +2930,7 @@ private void registerMacerator(int id) {
             name = TR_NAME(Type.NONE, "200V Macerator");
 
             MaceratorDescriptor desc = new MaceratorDescriptor(name,
-                "maceratorb", MVU, 400,// double nominalU,double nominalP,
+                "maceratorb", MVU, 200 * 5,// double nominalU,double nominalP,
```
jrddunbr: There are more magic fields than comments. Break out by comma and add comments.

src/main/java/mods/eln/Eln.java
```diff
@@ -2900,9 +2942,33 @@ private void registerMacerator(int id) {
...
+
+	private void registerArcFurnace(int id) {
+
+        int subId, completId;
+        String name;
+        {
+            subId = 0;
+            name = TR_NAME(Type.NONE, "800V Arc Furnace");
+
+            ArcFurnaceDescriptor desc = new ArcFurnaceDescriptor(
```
jrddunbr: This is exactly what I'm looking for above. This is great!

src/main/java/mods/eln/Eln.java
```diff
@@ -3738,6 +3830,18 @@ private void registerDust(int id) {
             Data.addResource(element.newItemStack());
             addToOre("dustCinnabar", element.newItemStack());
         }
+        {
+            id = 11;
+
+            name = TR_NAME(Type.NONE, "Resistive Dust");
+
+            element = new GenericItemUsingDamageDescriptorWithComment(name,// iconId,
```
jrddunbr: These comments don't make much sense.

src/main/java/mods/eln/Eln.java
```diff
@@ -3846,6 +3950,21 @@ private void registerIngot(int id) {
             Data.addResource(element.newItemStack());
             addToOre("quicksilver", element.newItemStack());
         }
+        {
+            subId = 9;
+            completId = subId + (id << 6);
+
+            name = TR_NAME(Type.NONE, "Silver Ingot");
+            element = new GenericItemUsingDamageDescriptorWithComment(name,// iconId,
```
jrddunbr: Comments are odd here too?

src/main/java/mods/eln/Eln.java
```diff
@@ -4642,6 +4781,239 @@ private void registerRawCable(int id) {
...
+            GenericItemUsingDamageDescriptor descriptor;
+            subId = 10;
+            completId = subId + (id << 6);
+            name = TR_NAME(Type.NONE, "Inert Canister");
+
+            descriptor = new GenericItemUsingDamageDescriptor(name);
+            sharedItem.addElement(completId, descriptor);
+            Data.addResource(descriptor.newItemStack());
+        }
+        {
```
jrddunbr: Don't add these until they do something.

src/main/java/mods/eln/Eln.java
```diff
@@ -4642,6 +4781,239 @@ private void registerRawCable(int id) {
...
+            subId = 15;
+            name = TR_NAME(Type.NONE, "Copper Folded Plate");
+            GenericItemUsingDamageDescriptorWithComment desc = new GenericItemUsingDamageDescriptorWithComment(
+                name, new String[]{"Now it can fit inside a bucket"});
```
jrddunbr: These are not being handled by the translation code properly.

src/main/java/mods/eln/Eln.java
```diff
@@ -5710,32 +6157,33 @@ private void recipeBattery() {
...
 
         addRecipe(findItemStack("Life Oriented Battery"),
-            "P P",
-            " B ",
-            "P P",
+            "PBP",
             'B', findItemStack("Cost Oriented Battery"),
             'P', new ItemStack(Items.gold_ingot));
+        addRecipe(findItemStack("Experimental Battery"),
+            " S ",
```
jrddunbr, line 6073: Use spaces not tabs.

jrddunbr: In fact, most of the stuff you are writing uses tabs. Stop that and start using 4 spaces. Please configure your editor to do this.

jrddunbr: You can use a feature called "soft tab" in most editors where it replaces pressing the tab key with 4 spaces.

src/main/java/mods/eln/Eln.java
```diff
@@ -6369,7 +6843,20 @@ private void recipeElectricalDrill() {
             'C', dictAdvancedChip,
             'M', findItemStack("Advanced Electrical Motor"),
             'P', new ItemStack(Items.diamond_pickaxe));
-
+		addRecipe(findItemStack("Turbo Electrical Drill"),
+            "RCR",
```
jrddunbr: Use spaces to indent again. Pretty much all of these indentation failures are caused by using tabs.

src/main/java/mods/eln/sixnode/resistor/ResistorContainer.java
```diff
@@ -18,9 +18,10 @@
 
     public ResistorContainer(EntityPlayer player, IInventory inventory) {
         super(player, inventory, new Slot[]{
-            new SlotFilter(inventory, coreId, 132, 8, 64, ItemStackFilter.OreDict("dustCoal"),
+            new SlotFilter(inventory, coreId, 135, 8, 64, ItemStackFilter.OreDict("dustResistive"),
```
jrddunbr, line 21: Magic numbers each need a line and comment.

src/main/java/mods/eln/transparentnode/electricalmachine/ArcFurnaceDescriptor.java
```diff
@@ -0,0 +1,78 @@
...
+
+import mods.eln.misc.*;
+import mods.eln.misc.Obj3D.Obj3DPart;
+import mods.eln.sim.ThermalLoadInitializer;
+import mods.eln.sixnode.electricalcable.ElectricalCableDescriptor;
+import net.minecraft.entity.item.EntityItem;
+import org.lwjgl.opengl.GL11;
+
+public class ArcFurnaceDescriptor extends ElectricalMachineDescriptor {
+    private float tyOn;
```
jrddunbr: what on earth are these.

.gitignore
```diff
@@ -24,3 +24,4 @@ options.txt
 # Blender backup files
 *.blend1
 *.blend2
+*.bat
```
jrddunbr: Didn't you say at one point that you didn't mean to do this and removed it on my fork? Just trying to jog some memory. I can't remember why we did or didn't want this here. Potentially it interfered with Gradle?

src/main/java/mods/eln/Eln.java
jrddunbr: (no offense intended, but if I had a penny for every time I made an indent comment, I'd have a candy bar :D )

jrddunbr (changes requested):
I've noticed that your editor is using tabs a lot. You should really configure the soft-tab functionality of your text editor to use 4 spaces instead. I have not commented everywhere that there is a tab or a failure to indent properly, but there are 341 tabs in this document that shouldn't be there (and most are probably not yours).

src/main/java/mods/eln/Eln.java
jrddunbr: (see most of the diff for a better view - basically, the indentation and comments are great)

jrddunbr:
Your code also failed the Travis build because you're not using the new functions Baugh set up for those pesky variables we had problems with a few weeks back.

.gitignore
OmegaHaxors: They're proper gitignores now, thank you.

src/main/java/mods/eln/Eln.java
OmegaHaxors: I have taken your feedback on the layout, but I will not add comment as that just tends to leave it a mess. Use your IDE to find out what each thing does, it's a lot cleaner.

src/main/java/mods/eln/Eln.java
OmegaHaxors: Caught automatically by refactor.

src/main/java/mods/eln/Eln.java
OmegaHaxors: It's a boolean which represents if it uses Life mechanics or not. You can find this by hovering over it with any competent IDE.

src/main/java/mods/eln/Eln.java
OmegaHaxors: This is a compatibility branch to the main repo, while it may be active on your build, this one's intended to go into the main build, which is why it needs to be defined.

src/main/java/mods/eln/Eln.java
OmegaHaxors: Thanks for catching that, it was a copy and paste. I went ahead and cleaned that up.

src/main/java/mods/eln/Eln.java
OmegaHaxors: They need localizing.

src/main/java/mods/eln/sixnode/resistor/ResistorContainer.java
OmegaHaxors: It was a copy/paste, won't be changing.

OmegaHaxors:
That should address any issue that were had.

OmegaHaxors:
I will need help with one bit though, the localization on the new cables seems to fail and i'm not quite sure why.

OmegaHaxors:
Confirmed now not to cause any regressions when updating.
That said, this feature pack won't contain any of the newer things (such as the arcmetal/clay blocks) so make sure you have those in there when building.

src/main/java/mods/eln/Eln.java
```diff
@@ -253,13 +253,30 @@
     public int electricalInterSystemOverSampling;
 
     public CopperCableDescriptor copperCableDescriptor;
+    public GraphiteDescriptor GraphiteDescriptor;
 
+    /*public ElectricalCableDescriptor creativeCableDescriptor;
```
Baughn: Don't comment out code, just delete it.

src/main/java/mods/eln/Eln.java
```diff
@@ -253,13 +253,30 @@
...
+    public ElectricalCableDescriptor T2TransmissionCableDescriptor;
+    public ElectricalCableDescriptor T1TransmissionCableDescriptor;*/
     public ElectricalCableDescriptor veryHighVoltageCableDescriptor;
```
Baughn, line 258: Change the 'normal' qualities to start with 'normalQuality'.

Also, do not start variable names with capital letters. That's for constants only.

src/main/java/mods/eln/Eln.java
```diff
@@ -595,6 +612,7 @@ public void preInit(FMLPreInitializationEvent event) {
         registerBattery(16);
         registerElectricalFurnace(32);
         registerMacerator(33);
+        registerArcFurnace(34);
```
Baughn: The arc furnace does not belong in this PR, especially since it isn't done.

src/main/java/mods/eln/Eln.java
```diff
@@ -791,12 +810,14 @@ public void load(FMLInitializationEvent event) {
         recipeMiningPipe();
         recipeTreeResinAndRubber();
         recipeRawCable();
+        recipeGraphite();
```
Baughn: Ditto its recipes.

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,134 +1173,167 @@ private void registerElectricalCable(int id) {
...
+        double nominalcablevoltage = 0;
+        double nominalcablepower = 0;
+
+        stdCableRender50V = new CableRenderDescriptor("eln", "sprites/cable.png", 1.95f, 0.95f);
+        stdCableRender200V = new CableRenderDescriptor("eln", "sprites/cable.png", 2.95f, 0.95f);
+        stdCableRender800V = new CableRenderDescriptor("eln", "sprites/cable.png", 3.95f, 1.95f);
+        stdCableRender3200V = new CableRenderDescriptor("eln", "sprites/cableVHV.png", 3.95f, 1.95f);
+
+        List<String> voltagelist = new ArrayList<>();
+        voltagelist.add("Low ");
```
Baughn:
```
List<String> voltageList = new ArrayList<>(Arrays.asList(
  "low",
  "medium",
  ...
);

```

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,134 +1173,167 @@ private void registerElectricalCable(int id) {
...
+        voltagelist.add("Medium ");
+        voltagelist.add("High ");
+        voltagelist.add("Very High ");
+        List<String> qualitylist = new ArrayList<>();
+        qualitylist.add("");
```
Baughn: Ditto

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,134 +1173,167 @@ private void registerElectricalCable(int id) {
...
+        qualitylist.add("Shoddy ");
+        qualitylist.add("Quality ");
+        qualitylist.add("Perfect ");
+        List<Double> valuelist = new ArrayList<>();
+        valuelist.add(1d);
```
Baughn: Ditto

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,134 +1173,167 @@ private void registerElectricalCable(int id) {
...
+        valuelist.add(0.5d);
+        valuelist.add(2d);
+        valuelist.add(4d);
+        List<String> renderlist = new ArrayList<>();
+        renderlist.add("sprites/cable.png");
```
Baughn: You get the picture.

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,134 +1173,167 @@ private void registerElectricalCable(int id) {
...
+        rendercablesizes.add(0.95f);
+        rendercablesizes.add(2.95f);
+        rendercablesizes.add(0.95f);
+        rendercablesizes.add(3.95f);
+        rendercablesizes.add(1.95f);
+
+        int renderspriteindex = 0;
+        int renderfloatindex = 0;
+        subId = 3;
+        while (cablevoltage <= 3) {
```
Baughn: Use a nested for-loop instead.

src/main/java/mods/eln/Eln.java
```diff
@@ -4642,6 +4756,238 @@ private void registerRawCable(int id) {
         }
     }
 
+    private void registerArc(int id) {
```
Baughn: The arc furnace should be a different PR. Or at least needs to be finished before I'll accept this one.

src/main/java/mods/eln/Eln.java
```diff
@@ -5790,6 +6228,35 @@ private void recipeGridDevices(HashSet<String> oreNames) {
             'C', findItemStack("Optimal Ferromagnetic Core"),
             'T', findItemStack("DC-DC Converter")
         );
+
+        // I don't care what you think, if your modpack lacks steel then you don't *need* this much power.
+		// Or just use the new Arc furnace. Other mod's steel methods are slow and tedious and require huge multiblocks.
```
Baughn: Whitespace error.

src/main/java/mods/eln/Eln.java
```diff
@@ -5813,13 +6280,17 @@ private void recipeGridDevices(HashSet<String> oreNames) {
 
 
     private void recipeElectricalFurnace() {
-        addRecipe(findItemStack("Electrical Furnace"),
+        /*addRecipe(findItemStack("Electrical Furnace"),
```
Baughn: Why are you disabling the furnace?

src/main/java/mods/eln/Eln.java
```diff
@@ -6404,20 +6903,86 @@ private void recipeTreeResinAndRubber() {
...
+            "III",
+            'I', findItemStack("Arc Metal Ingot")); //this should be oredict steel, but I can't figure that out. Plz to help??
+        addRecipe(findItemStack("T2 Transmission Cable", 6),
+            "III",
+            'I', findItemStack("Arc Clay Ingot")); //likewise, but with aluminium */
 
     }
 
+    private void recipeGraphite() {
+        /*addRecipe(findItemStack("Creative Cable", 1),
```
Baughn: Again, don't comment out code. Just delete it.

Baughn (changes requested):

OmegaHaxors:
EDIT: Alright that should be everything taken care of. Just have to test everything and make sure there isn't any issues, then I will push the update and it will be ready for merging.

OmegaHaxors:
Thank you for being patient, that should be everything.

src/main/java/mods/eln/Eln.java
Baughn: I note that these variables still start with capital letters. Let me be more specific: Use the alpacaCase pattern, e.g. `lqVeryHighVoltageCableDescriptor`.

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,134 +1165,161 @@ private void registerElectricalCable(int id) {
...
+        double power = 0;
+        double nominalcablevoltage = 0;
+        double nominalcablepower = 0;
+
+        stdCableRender50V = new CableRenderDescriptor("eln", "sprites/cable.png", 1.95f, 0.95f);
+        stdCableRender200V = new CableRenderDescriptor("eln", "sprites/cable.png", 2.95f, 0.95f);
+        stdCableRender800V = new CableRenderDescriptor("eln", "sprites/cable.png", 3.95f, 1.95f);
+        stdCableRender3200V = new CableRenderDescriptor("eln", "sprites/cableVHV.png", 3.95f, 1.95f);
+
+        List<String> voltagelist = new ArrayList<>(Arrays.asList(
```
Baughn: Nit: You don't need to have 'list' in the names; that's supplied by the type. "voltages" would be fine.

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,134 +1165,161 @@ private void registerElectricalCable(int id) {
...
+            2.95f,
+            0.95f,
+            3.95f,
+            1.95f));
+
+        int renderspriteindex = 0;
+        int renderfloatindex = 0;
+        subId = 3;
+        for(int cablevoltage = 0; cablevoltage <=3; cablevoltage++){
+        for(int cablequality = 0; cablequality <=3; cablequality++) {
```
Baughn: Indentation.

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,134 +1165,161 @@ private void registerElectricalCable(int id) {
...
+            //while (cablevoltage <= 3) {
```
Baughn: Don't *add* commented-out code.

src/main/java/mods/eln/Eln.java
Baughn: Add a comment about what the indexes mean.

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,134 +1165,161 @@ private void registerElectricalCable(int id) {
...
+            "Standard ",
+            "Shoddy ",
+            "Quality ",
+            "Perfect "));
+        List<Double> valuelist = new ArrayList<>(Arrays.asList(
+            1d,
+            0.5d,
+            2d,
+            4d));
+        List<String> renderlist = new ArrayList<>(Arrays.asList(
```
Baughn: Especially here, since it obviously isn't the same as above.

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,134 +1165,161 @@ private void registerElectricalCable(int id) {
...
+        for(int cablevoltage = 0; cablevoltage <=3; cablevoltage++){
+        for(int cablequality = 0; cablequality <=3; cablequality++) {
+            //while (cablevoltage <= 3) {
+            renderspriteindex = 0;
+            subId++;
+            name = TR_NAME(Type.NONE, qualitylist.get(cablequality) + voltagelist.get(cablevoltage) + "Voltage Cable");
+
+            renderspriteindex = cablequality;
+
+            switch (cablevoltage) {
```
Baughn: I'm very tempted to say "Do this in Kotlin", because the code would be much cleaner. I won't, but I will at some point in the future, if not about this code; you need to learn it anyway.

In particular, read about 'when' expressions [here](http://kotlinlang.org/docs/reference/control-flow.html), right now. You'll love them, and there's no way to forget a break. Switches should be limited to situations where you *want* to leave out breaks.

src/main/java/mods/eln/Eln.java
```diff
@@ -1078,6 +1091,7 @@ public void onServerStarting(FMLServerStartingEvent ev) {
     public CableRenderDescriptor stdCableRender200V;
     public CableRenderDescriptor stdCableRender800V;
     public CableRenderDescriptor stdCableRender3200V;
+    public CableRenderDescriptor activerendermodel;
```
Baughn: Why is this an instance-wide variable?

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,134 +1165,161 @@ private void registerElectricalCable(int id) {
...
+                    break;
+                case 3:
+                    power = VVP();
+                    nominalcablevoltage = VVU;
+                    renderfloatindex = 4;
+                    renderspriteindex = renderspriteindex + 4;
+                    break;
+            }
+            activerendermodel = new CableRenderDescriptor("eln", renderlist.get(renderspriteindex), rendercablesizes.get(renderfloatindex), rendercablesizes.get(renderfloatindex + 1));
+            desc = new ElectricalCableDescriptor(name, activerendermodel, "Haxors was Here", false);
```
Baughn: Cute, but no.

src/main/java/mods/eln/Eln.java
```diff
@@ -1148,134 +1165,161 @@ private void registerElectricalCable(int id) {
...
+                            meduimVoltageCableDescriptor = desc;
+                            break;
+                        case 2:
+                            highVoltageCableDescriptor = desc;
+                            break;
+                        case 3:
+                            veryHighVoltageCableDescriptor = desc;
+                            break;
+                    }
+                case 1:
```
Baughn: Fair warning, I'll probably rewrite all of this in the future. Because it could be much cleaner... in Kotlin. :P

src/main/java/mods/eln/Eln.java
```diff
@@ -1339,25 +1383,31 @@ private void registerBattery(int id) {
         double stdDischargeTime = 4 * 60;
         double stdU = LVU;
         double stdP = LVP() / 4;
-        double stdEfficiency = 1.0 - 2.0 / 50.0;
+        double stdEfficiency = 1.0 - 2.0 / 50.0; //96%
```
Baughn: Just make it "0.96". Having the computation doesn't make this any clearer.

src/main/java/mods/eln/Eln.java
```diff
@@ -2692,12 +2752,12 @@ private void registerHeatFurnace(int id) {
             name = TR_NAME(Type.NONE, "Stone Heat Furnace");
 
             HeatFurnaceDescriptor desc = new HeatFurnaceDescriptor(name,
-                "stonefurnace", 1000,
+                "stonefurnace", 4000,
```
Baughn, line 2776: You need to describe this change in the PR description. Actually, everything really.

src/main/java/mods/eln/Eln.java
```diff
@@ -2729,7 +2789,7 @@ private void registerTurbine(int id) {
             name = TR_NAME(Type.NONE, "50V Turbine");
             double RsFactor = 0.1;
             double nominalU = LVU;
-            double nominalP = 300 * heatTurbinePowerFactor;
+            double nominalP = 1000 * heatTurbinePowerFactor; // it was 300 before
```
Baughn: No need for the comment.

src/main/java/mods/eln/Eln.java
```diff
@@ -2743,7 +2803,7 @@ private void registerTurbine(int id) {
             name = TR_NAME(Type.NONE, "200V Turbine");
             double RsFactor = 0.10;
             double nominalU = MVU;
-            double nominalP = 500 * heatTurbinePowerFactor;
+            double nominalP = 2000 * heatTurbinePowerFactor; // it was 2000 before
```
Baughn: Ditto.

src/main/java/mods/eln/Eln.java
```diff
@@ -2888,20 +2948,19 @@ private void registerMacerator(int id) {
             name = TR_NAME(Type.NONE, "200V Macerator");
 
             MaceratorDescriptor desc = new MaceratorDescriptor(name,
-                "maceratorb", MVU, 400,// double nominalU,double nominalP,
-                MVU * 1.25,// double maximalU,
-                new ThermalLoadInitializer(80, -100, 10, 100000.0),// thermal,
-                meduimVoltageCableDescriptor,// ElectricalCableDescriptor
-                // cable
+                "maceratorb", MVU, 1000,
```
Baughn: Also this. The change itself is fine, but it needs to be included in the descripotion.

src/main/java/mods/eln/Eln.java
```diff
@@ -4679,7 +4833,7 @@ private void registerBrush(int id) {
 
             addShapelessRecipe(emptyStack.copy(),
                 new ItemStack(Blocks.wool, 1, idx),
-                new ItemStack(Items.iron_ingot));
+                findItemStack("Iron Cable"));
```
Baughn: ..fair enough, but I feel like it'll flop around in my hand. Maybe change the recipe to produce multiple outputs instead; 1 wool is actually quite a lot of wool, given 3 makes a bed.

src/main/java/mods/eln/Eln.java
```diff
@@ -4955,7 +5109,7 @@ private void registerMiscItem(int id) {
         {
             subId = 23;
             GenericItemUsingDamageDescriptorWithComment desc = new GenericItemUsingDamageDescriptorWithComment(
-                TR_NAME(Type.NONE, "Advanced Machine Block"), new String[]{}); // TODO: Description.
+                TR_NAME(Type.NONE, "Advanced Machine Block"), new String[]{});
```
Baughn: Don't remove TODOs without doing what they say, when what they say to do is quite reasonable. Just add the description. :P

src/main/java/mods/eln/Eln.java
```diff
@@ -6639,9 +6912,31 @@ private void recipeMacerator() {
             new ItemStack[]{new ItemStack(Blocks.gravel)}, 1.0 * f));
         maceratorRecipes.addRecipe(new Recipe(new ItemStack(Blocks.gravel),
             new ItemStack[]{new ItemStack(Items.flint)}, 1.0 * f));
-
         maceratorRecipes.addRecipe(new Recipe(new ItemStack(Blocks.dirt),
             new ItemStack[]{new ItemStack(Blocks.sand)}, 1.0 * f));
+        double batf = 50000 * batteryCapacityFactor + f; //to prevent overunity
```
Baughn: How does it do so?

"Prevent overunity by using more power than a new battery stores."

src/main/java/mods/eln/transparentnode/electricalmachine/ArcFurnaceDescriptor.java
Baughn: Yes, please explain. By comment.

src/main/java/mods/eln/transparentnode/turret/TurretSlowProcess.java
```diff
@@ -18,6 +18,10 @@
 import java.util.List;
 import java.util.Random;
 
+import static scala.Console.print;
```
Baughn: No. We don't use Scala.

Baughn (changes requested):

src/main/java/mods/eln/Eln.java
```diff
@@ -1187,7 +1187,7 @@ private void registerElectricalCable(int id) {
             1d,   //Normal
             0.5d, //Half
             2d,   //Double
-            4d)); //Quadruple
+            10d)); //Ten Times
```
Baughn: Nit: "Dectuple"

src/main/java/mods/eln/Eln.java
```diff
@@ -5279,10 +5279,10 @@ private void recipeElectricalCable() {
 
         addRecipe(lowVoltageLQCableDescriptor.newItemStack(12), //Shoddy Low Voltage Cable
             "WWW",
-            "GCG",
+            "DCD",
             "WWW",
             'C', "ingotCopper",
-            'G', new ItemStack(Blocks.gravel),
+            'D', new ItemStack(Blocks.dirt),
```
Baughn: That would just crumble. Granted, so would the dirt. How about clay?

src/main/java/mods/eln/Eln.java
```diff
@@ -6885,14 +6885,14 @@ private void recipeMacerator() {
             new ItemStack[]{findItemStack("Coal Dust", 1)}, 1.0 * f));
         maceratorRecipes.addRecipe(new Recipe(new ItemStack(Items.coal, 1, 1),
             new ItemStack[]{findItemStack("Coal Dust", 1)}, 1.0 * f));
-        maceratorRecipes.addRecipe(new Recipe(new ItemStack(Blocks.sand, 1),
+        maceratorRecipes.addRecipe(new Recipe(new ItemStack(Items.flint, 4),
```
Baughn, line 7000: ...fair enough.

src/main/java/mods/eln/Eln.java
```diff
@@ -6905,12 +6905,12 @@ private void recipeMacerator() {
         maceratorRecipes.addRecipe(new Recipe(findItemStack("Tungsten Ingot"),
             new ItemStack[]{findItemStack("Tungsten Dust", 1)}, 0.5 * f));
 
-        maceratorRecipes.addRecipe(new Recipe(new ItemStack(Blocks.cobblestone),
-            new ItemStack[]{new ItemStack(Blocks.gravel)}, 1.0 * f));
+        maceratorRecipes.addRecipe(new Recipe(new ItemStack(Blocks.cobblestone,32),
+            new ItemStack[]{new ItemStack(Blocks.gravel)}, 10.0 * f));
```
Baughn: This is a massive downgrade -- 32 cobblestone for 1 gravel. No, cobblestone is not _that_ common, and I don't usually build a cobblegen as the first thing I do.

src/main/java/mods/eln/Eln.java
```diff
@@ -6905,12 +6905,12 @@ private void recipeMacerator() {
...
         maceratorRecipes.addRecipe(new Recipe(new ItemStack(Blocks.gravel),
-            new ItemStack[]{new ItemStack(Items.flint)}, 1.0 * f));
-        maceratorRecipes.addRecipe(new Recipe(new ItemStack(Blocks.dirt),
-            new ItemStack[]{new ItemStack(Blocks.sand)}, 1.0 * f));
+            new ItemStack[]{new ItemStack(Items.flint)}, 0.25 * f));
+        maceratorRecipes.addRecipe(new Recipe(new ItemStack(Blocks.stone),
+            new ItemStack[]{new ItemStack(Blocks.sand)}, 5.0 * f));
```
Baughn: I think it'd be better to use the mod-typical chain: Stone to cobble to sand, or alternately stone to cobble to gravel to sand.

As for earth, it would actually make physical sense to allow macerating earth to clay. Well, some physical sense. More than having any other block as input to that.

src/main/java/mods/eln/Eln.java
OmegaHaxors: Reverted based on feedback.

src/main/java/mods/eln/Eln.java
OmegaHaxors: I'm thinking of just getting rid of the 'filler' item so that it only costs 1 ingot and some wood

OmegaHaxors:
Alright I think that's everything.

.gitignore
```diff
@@ -14,6 +14,8 @@ eclipse
 options.txt
 /out
 /classes
+/build-play.bat
+/build-full.bat
```
Baughn: Nit: Really, just make this *.bat.

.gitignore
```diff
@@ -24,3 +26,4 @@ options.txt
 # Blender backup files
 *.blend1
 *.blend2
+
```
Baughn: Nit: Don't add extra blank lines at the end.

src/main/java/mods/eln/transparentnode/turret/TurretSlowProcess.java
```diff
@@ -206,7 +207,11 @@ else if (element.getTurretAngle() <= -element.getDescriptor().getProperties().ac
                         }
 
                     if (visible) {
-                        element.play(new SoundCommand("eln:TurretFire").mulVolume(0.4));
+                        if(entity.getHealth()>0) {
```
Baughn: Nit: Leave spaces around the >

src/main/java/mods/eln/transparentnode/turret/TurretSlowProcess.java
```diff
@@ -237,7 +242,7 @@ public void enter() {
 
         @Override
         public State state(double time) {
-            if (target.isDead) return new SeekingState();
+            if (target.getHealth()<=0) return new SeekingState();
```
Baughn: Ditto

src/main/java/mods/eln/transparentnode/turret/TurretSlowProcess.java
```diff
@@ -319,14 +324,17 @@ public ShootState(EntityLivingBase target) {
...
+		    target.attackEntityFrom(new DamageSource("Unknown"), 5);
+			element.shoot();
             element.play(new SoundCommand("eln:LaserGun"));
+		    }
         }
 
         @Override
         public State state(double time) {
-            if (target == null || target.isDead)
+            if (target == null || target.getHealth()<=0)
```
Baughn: Ditto

Baughn (changes requested):
Also, you really need to update the PR description to describe everything that's in it.

PR #905, opened 2019-03-28
Arc furnace sound

keever50:
An alternative sound for the arc furnace


Baughn (dismissed):
Seems fine to me, though obviously we should make it sound more like https://www.youtube.com/watch?v=G6Uxh-xtU-g... maybe? XD

Maybe not.

keever50:
Yeah sure. I dont mind changing sound at all. I can try something more industrial like that. I avoided this because the most you are hearing is the cooling fans and expanding/exploding metal. I can add rumble and melting metal sound if you want ^^

Baughn:
Some sounds from the melting metal would be neat. Low-frequency, probably; it sounds better for what's the some of the heaviest machinery in the mod, and bass is also easier to listen to for long periods.

Don't forget it needs to loop.

keever50:
Should be fine now ^^

PR #923, opened 2019-04-15
BugFix

patrudaz:
- Multiple writes works now even if the connection stay open


Dolu1990 (approved):
Ganz schön, danke viele mal ^.^

PR #945, opened 2019-07-24
Sanity

jrddunbr:
This pull request cleans up the code in Node a bit. Removing excess white space that makes classes hard to read on one screen, changing constants to uppercase, moving variables to the top of the class, compressing catch statements,... typical stuff.


jrddunbr:
Codacy needs to cool off a bit on that documentation check. It's never going to get it.

You can configure that in the settings for the repository about what to care about.

PR #948, opened 2019-10-31
Update translated languages files.

metc:
Translations are pulled manually from Transifex. This can be done using
the Transifex CLI client using the `tx pull -a -f` command.

We should automate this with CI and a cron job. Creating a Docker image with Java build tools, Python and the Transifex client would be the first step.

We have different lang files for the same "base" language, like `es.lang`, `es_419.lang`, and `es_AR.lang`. I do not think they really improve the translations. We should think about merging those translations.


jrddunbr:
Nice. I wasn't really sure how we pull those in, and I've been having people submit the language file itself for my fork, because it was so unclear how to do stuff on Transifex. Perhaps some instructions on the Wiki would be a useful thing, for submitting translations at least, but also perhaps how to apply them in the source code.

metc:
> Nice. I wasn't really sure how we pull those in, and I've been having people submit the language file itself for my fork, because it was so unclear how to do stuff on Transifex. Perhaps some instructions on the Wiki would be a useful thing, for submitting translations at least, but also perhaps how to apply them in the source code.

Lack of documentation, yes... CI uses `https://github.com/Electrical-Age/ElectricalAge/blob/develop/script/transifex.sh` to push the source language file to Transifex.
We do not have an automated way of pulling the translated files from Transifex yet. This is a manual process we usually do before every releases.

jrddunbr:
Hmm. I see why automating that would be a good idea. Plus, we haven't had a release in months because port work is slow and I haven't started the massive task of PR'ing my branches changes.

